PR #9, opened 2021-01-05
Add support for Oneplus Dialer

amine250:
Hope this is the right way to add support for the Oneplus dialer, if not, I'll be happy to fix it.


Donnnno:
Yes! That's the right way! I'll merge it right away for the next update :)

PR #10, opened 2021-01-05
added missing entry for SuperFreezz

0slav:


0slav:
Oh, I should probably leave some info. The icon was added and looks great, just this entry slipped out somehow along the way.

Donnnno:
Thanks! Probably forgot to add it in there, oops.

PR #14, opened 2021-01-05
Add Google Pay icon

amine250:
This icon was a PITA to make while using only lines 😄. 
So, I didn't have a choice but to use a shape of width 1pt for the 'y' so I could have the same character as the official icon.

It turns out the icon is slightly different but we can't make it exactly similar without violating the 1pt rule.


Donnnno:
I'll take a look at it, but I'm not a huge fan of using text in icons.

I also noticed you're using an old Google Pay logo, do you mind if we use this one? (I can create it pretty easily)

image

Oh, and I'm working on a font for this pack, it's almost finished, but it should be pretty easy to use in the future if you'll need a text icon. :)

Thanks!

amine250:
I wasn't aware there's new GPay icon, and that there are even 2 packages of it on the play store:

- old: https://play.google.com/store/apps/details?id=com.google.android.apps.walletnfcrel
- new: https://play.google.com/store/apps/details?id=com.google.android.apps.nbu.paisa.user

That's fine if you can recreate it as the new one.

And I also can't wait for the font, it'll make matters easier 😄. 

Thanks.

Donnnno:
Oh, that's interesting haha.

Here's a preview of the font: many icons already use the letters, I just have to finish up some last things :) 
image

Donnnno:
image

amine250:
@Donnnno, did you add the other package/activity name to the appfilter.xml ?
The package name is com.google.android.apps.nbu.paisa.user but I don't know the activity name.

Donnnno:
Nah, I don't have the activity name, if someone misses it, they can make a request ;)

PR #15, opened 2021-01-06
Assign OpenScale icon to Lose It

studio315b:
Nobody will be using both, and the icons are close enough. Is it perfect? No, but not every launcher supports arbitrary icons.


Donnnno:
Cool, I'll add it to the icon pack :)

PR #18, opened 2021-01-07
fix gradle checksum

TBog:
check https://gradle.org/release-checksums/

If the checksum does not match Android Studio may crash. 
It does in my case.


Donnnno:
Oh, yea I had some problems with that too, I didn't really know what to do about that, thanks!

PR #20, opened 2021-01-07
Improve IconActivity

TBog:
This PR will make a grid layout for IconActivity

* This will fix the vector loading and allow dynamic number of columns in the grid.
* You can change the size of the icons from `dimens.xml` > `icon_preview_size` right now it's 48dp
* The icons are loading when needed, not like before where all got loaded at the start
* ~~The decoding of the icon resource is done by the ImageView when calling `icon.setImageResource(resId);`~~ 
* The icon decoding is done asynchronously off the UI thread. A Drawable is returned on the UI thread to be associated with the ImageView. 

If you need any tweaks please leave a comment.

### DIFF
--- a/app/src/main/java/com/donnnno/arcticons/activities/IconActivity.java
+++ b/app/src/main/java/com/donnnno/arcticons/activities/IconActivity.java
@@ -1,127 +1,67 @@
 package com.donnnno.arcticons.activities;
 
-import android.graphics.Rect;
 import android.os.Bundle;
 import android.text.Editable;
-import android.text.InputType;
 import android.text.TextWatcher;
-import android.view.Gravity;
-import android.view.View;
-import android.widget.ImageView;
-import android.widget.LinearLayout;
-import android.widget.ScrollView;
 import android.widget.EditText;
+import android.widget.GridView;
 
 import com.donnnno.arcticons.R;
-import com.donnnno.arcticons.utils.ImageUtils;
-import com.donnnno.arcticons.utils.ScreenUtils;
+import com.donnnno.arcticons.adapters.IconAdapter;
 
 import java.util.ArrayList;
-import java.util.Arrays;
+import java.util.Collections;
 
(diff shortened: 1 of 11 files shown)


Donnnno:
I'll take a look at it, thanks!

Donnnno:
Wow, that works great, way less laggy than before haha. 

I have only two things to that come to my mind right now: 
- Can you change the background to black?
- Maybe a long press icon name thing would be nice to have, but not necessary 

Thanks!

TBog:
> - Can you change the background to black?
This can be done from the xml layout easily 

> - Maybe a long press icon name thing would be nice to have, but not necessary 
I wanted to add it but I was not sure if you desired it as it was not implemented. Is a [Toast](https://developer.android.com/guide/topics/ui/notifiers/toasts) enough? It should display the drawable name, right?

Donnnno:
> Is a Toast enough?

Yes, that was exactly what I was thinking about!

TBog:
* We can also share the image if you think that would be useful.
  * We could also implement a content resolver so requests for images could show this app

Donnnno:
> We can also share the image if you think that would be useful

Useful for what exactly? 

> We could also implement a content resolver so requests for images could show this app

Do you mean like an icon request button or something?

TBog:
> > We can also share the image if you think that would be useful
> 
> Useful for what exactly? 

Share the icon to WhatsApp or Snapseed 

> > We could also implement a content resolver so requests for images could show this app
> 
> Do you mean like an icon request button or something? 

When another app requests an image, like the browser when you press Choose Image on a website

Donnnno:
Ah okey, I think we don't need a share button for the icons. The second thing could be a nice to have but also not necessary :^)

Donnnno:
Looks great! Is it ready to merge?

TBog:
> Looks great! Is it ready to merge?

It's ready to merge.

Donnnno:
Cool! thanks :)

PR #19, opened 2021-01-07
Vector drawable example

TBog:
This is an example to fix #16 for the calculator icon
Steps:
1. Set project grouping to "Android"  
image

2. Right click on calculator and `New` > `Vector Asset`
image

3. Configure vector asset:
  * Asset Type: **Local file**
  * Name: **calculator**
  * Path: path to **Arcticons/icons/calculator.svg**
  * Size: 48dp x 48dp
image

4. Press **Next** and confirm path 
image

5. Now you should have 6 calculator entries 
image

6. Delete all `calculator.png` files 
image

PR #25, opened 2021-01-08
Removing generated files from repo and adding Makefile

jorants:
I love the icons and really wanted to use them on my eBook reader, but for this they had to be black instead of white. 
In the process of trying to change this I needed an easier way to regenerate all the PNGs. 

I removed all generated files (the XML files and PNG files) form my repo as I think it is cleaner not the have a lot of generated (binary) files in the github while the source is also present. I merged versions of the scripts in `other` into a makefile that compile everything that needs recompiling. This hopefully makes it easier for others to tinker as well.

A few perks of this setup:
 - Make knows what needs recompiling, so after a first run only new or changed icons get recompiled.
 - `make -j...` does a parallel compilation of the PNG files.
 - It is possibly clearer and easier for icon contributors. You can now simply add to `icons` for your PR, as make will detect the new one. This also allows contributors to compile their PR before submitting.

I also moved the appfilters to a (non-xml) file in the main directory, making it easier for people to add to it. This file is then converted on compilation. Finally, I added a section on building the project to the README. 

I understand it is a lot of changes, I hope you agree with some of them.


Donnnno:
Hi this looks interesting, we're currently working on changing the icons to xml over here: #19

Donnnno:
Wanted to test this build with android studio, but I'm getting this error: 
```
Arcticons\app\src\main\java\com\donnnno\arcticons\activities\IconActivity.java:24: error: cannot find symbol
        mImages = getResources().getStringArray(R.array.icon_pack);
                                                 ^
  symbol:   variable array
  location: class R
```

Donnnno:
Closing this because it isn't relevant anymore with the new dashboard and improved scripts.

PR #26, opened 2021-01-08
Add 4 more icons

0slav:
Cleanup may be necessary again.


Donnnno:
Cool, they look very good!

PR #39, opened 2021-01-13
Add ICBC and Night Screen

aaronhatesregex:


aaronhatesregex:
Knows how to deal with transform elements finally! I'll make another PR with all those icons later.

PR #45, opened 2021-01-16
Add 6 new icons (including missing mpv icon)

aaronhatesregex:


Donnnno:
Thanks, they look good!
One thing to keep in mind for the next time is consistent line weighs and try to use as less unnecessary shapes as possible (like big circles).

aaronhatesregex:
Thanks, I suspect I forgot to uncheck the `Move patterns along with the objects` option when I resize some patterns. I'll take care in the future and improve existing ones if I manage to find some time.

PR #52, opened 2021-01-29
Icons, XML update, minor tweaks.

RuffaWuffa:
Here's an additional 80+ icons and the updated XML files.
Should be possible to run raster.sh and then build.  Once anywhere-.svg from your recent update has it's name fixed.
Local build tested.


Donnnno:
Damn, that's a huge pull request, going to add it when I have some time. Thanks :)

RuffaWuffa:
I think that should take care of the appfilter.xml.  I've merged your latest commits into my branch and matched up the formatting inconsistencies with upstream.  The pull is showing no conflicts now.

I would recommend keeping with a consistent format for the entries in the XML files.  It can greatly simplify merging pull requests.
This may be because I used the included scripts to automate generating the three XML files.
I assume you're manually modifying the files instead of using the scripts.

The things that made the merge more difficult were:
- Spaces on blank lines.
- Some lines have a space before the /> and some do not.  The script is set up to put a space there, but the IconRequest XML doesn't have one.  Your choice as to which way to standardize that one.
- Indentation differences.  Just one line though, not indented at all..
- Trailing spaces at the end of a few lines.
- Sorting differences.  This may unavoidable; even the script may sort differently if the system language is different.

PR #58, opened 2021-02-14
Add BBT U, Tinfoil, Nintype

austenadler:
Adds icons:
- Keyboard 69
- Tinfoil for Facebook
- U by BB&T

### DIFF
--- a/app/src/main/res/xml/appfilter.xml
+++ b/app/src/main/res/xml/appfilter.xml
@@ -1725,6 +1725,9 @@
     <!-- Kernel Adiutor -->
     <item component="ComponentInfo{com.grarak.kerneladiutor/com.grarak.kerneladiutor.MainActivity}" drawable="adiutor" />
 
+    <!-- Keyboard 69 -->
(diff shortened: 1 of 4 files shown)


Donnnno:
Thanks, looks good! 

A thing to keep in mind for the next time:
- sizing, take a look at the template files for correct sizing
- use round caps & round joins

I've fixed those things for now ;)

austenadler:
Thank you for the pointers. What was the issue was the sizing? I am going to submit a pull with more icons so I want to get the sizing right. For Nintype/keybaord69:

|  | Doc size | Logo size | Logo x,y |
|-|-|-|-|
| Square Template | 48x48 | 38x38 | 5,5 |
| Submitted | 48x48 | 33.58x24.85 | 7.051,10.06 |
| Current main | 48x48 | 40.065x29.610 | 3.995,9.22 |

I used the square template as a general guide but I guess I made the logo too small. Would it be a safe rule to say ~4-8px on each side of the longest dimension for future logos?

Donnnno:
Hi, sizing can be really tough to do right. If you follow the basic shapes you're almost there.

I recommend to make nothing in its width or length larger than 39px except for circles.

PR #59, opened 2021-02-14
added Detoxdroid, Orange VMM ; corrected Zulip, Mumla, Attestation de deplacement

fanfu:


Donnnno:
Looks good, thanks!
The detoxdroid icon is adorable btw :)

PR #80, opened 2021-02-26
More icons

RuffaWuffa:
A slightly smaller collection of icons this time.  Less than 50.
This time they are mostly in the games category.
This concludes the list of apps and games I actually have installed.  Contributions will likely be smaller and less frequent after this.

Looking at the diff as I type this up makes it look like I added the appfilter.xml entry for Uber Eats.  Not sure how that happened.


Donnnno:
Hi, thanks again! Can you remove the updated `iconpack.xml` & `drawable.xml` ? I generate those automaticly when I push an update.

RuffaWuffa:
That should do it.  Rolled back the two xml files to the commit before I updated them.

Before doing a pull request I usually do a test build of the project and revert the build results before submitting.  I can just add those two files to the list to clean up in the future.

Donnnno:
Ah cool, thanks :)

Donnnno:
Hi, what did you change about the `xml.sh`?

Donnnno:
Hi, just took a more in-depth look at the icons, there were many things I forgot to mention with your last pull-request, but it was quite some work to fix it.
- Your documents are in points (pt) instead of **pixels (px)**
- caps & corners should have **round joins & caps**
- Many lines aren't connected with each other, that looks kinda messy. See the example below. 
- Sizing is weird, sometimes too small, like for example the `mobmuplat` icon. The shape is a circle which should have a size of 43x43px (see the templates for the default sizes), but right now it's 38,9087pt witch is quite smaller.
- Icons aren't centered correctly. Make sure to align the entire shape, so it is in the middle (x24,y24)

I hope you understand these points of critiques and can fix it, thanks.

Oh, also I can't merge stuff right now due to an error in the `appfilter.xml`.

Examples:
image
image
image

RuffaWuffa:
That is some useful info.  I was initially fitting most of the icons into the square template.  I had been targeting proper overlap within the stroke width instead of focusing on alignment of the stroke centerlines.  Wasn't aware of specific linecap and linejoin requirements.

The issues have been gone over and corrected as well as some other issues I discovered during the process.
Apparently an issue with Inkscape when adjusting multiple objects with different line properties; will for example, show round linecap selected if the majority of objects have it set, but require changing to a different linecap and back to rounded before the remaining ones get set properly.  Also found that some of the line connection issues were caused by using one object to cut an overlapping object.
A couple of the icons that have a hexagonal shape, I have adjusted to be close to fitting within the Circle template.  Also I have enlarged some of the narrow icons so as to be a closer match with either the rectangle vertical or rectangle horizontal.
Icons have been aligned as overall centre instead of loose or shape specific.

The pt vs px issue is confusing.  The majority of these icons started from your square template, so should inherit properties from that.
According to Inkscape all settings that I'm aware of are in px ( display units, page size, grid units, stroke width ).
Things that may be making a difference:  
- I'm working with a display having a 0.15875mm dot pitch ( 160 dpi ).
- The svg files content reads stroke width as stroke-width:1; which according to the svg standard should be equivalent to 1px line width, unless an unit typeis specified for the document or a differing viewport is being set.  Is this being misinterpreted in software somewhere?  Non-standard locked dpi is a known Adobe Illustrator problem, for example.
Attached here are three versions of a single svg file for comparison to see if one of the solutions I've thought of work out.
[svg test files.zip](https://github.com/Donnnno/Arcticons/files/6065502/svg.test.files.zip)

xml.sh:  I will be removing my changes to xml.sh.  They are unnecessary if you are automatically generating the files during updates on your end.  Using original script when doing temporary local builds doesn't cause me any problems.
The changes were to work around sorting issues which would made merging the drawable and iconpack xml files more complicated.  Automatic sorting is not consistent across different language settings, and when I ran the script I would get different sorting than the original files.

Donnnno:
Thank you for your hard work! It looks much cleaner now, great work! 👍 

> The pt vs px issue is confusing.

You're right! Those sizes were in pt, so that was my fault. I'll fix this right away.

Donnnno:
Ignore the px pt thing, it isn't really important. Great icons :)

PR #83, opened 2021-02-27
New icons

Lelio-Brun:
Hello, this PR adds some new icons to the icon set.
I'm a beginner with Inkscape so some of them might not be very well designed.

The activitie names appears in `appfilter.xml` already.


Donnnno:
Hi, thanks for your contribution! They look really good :) A couple of things:
- You have canged a lot with the appfilter.xml. Removing and adding lines of code, is there a reason for that?
- Some icons already exist: auditor, pdf viewer & slack, can you remove those?
- And can you make it so that all icons have round caps & round joins?

Lelio-Brun:
- Hmm I guess it results from the use of `sort.sh`, I shouldn't have used it
- Ok, sorry I didn't see them before!
- I tried to add round caps and joins in each icons, I hope it works

Donnnno:
Ah okey, thanks :)
Can you change the appfilter back to the master branch one and add it without using sort? Right now I can't add it because it has some conflicts.  :s

Lelio-Brun:
Ok, should be good now!

Donnnno:
Yes! Thanks, :) it looks good now.

PR #85, opened 2021-02-28
Slack alternative activity name

Lelio-Brun:
I understood why I initially included Slack in PR #83, it's because it seems the activity name is different.


Donnnno:
Thanks!

PR #88, opened 2021-03-01
Dynamic calendar support!!!!

Donnnno:
If you think that icon pack innovation is over, guess what... I still got more tricks:
- Dynamic calendar icons are here, different icons for every day of the month if your launcher supports it. Don't worry it still has a fallback for launchers that don't support it.
- updated xml.sh's way of sorting with numbers.

i n n o v a t i o n


Donnnno:
A preview image :)

image

Issue #81, opened 2021-02-26
Icon Request


MonsterSwagger:
[20210226_094013.zip](https://github.com/Donnnno/Arcticons/files/6048561/20210226_094013.zip)

Rhodney:
@Donnnno  As you already have several requests, I also take this one. I will do that in the next few days.

Donnnno:
@Rhodney thanks!! I've already applied the existing icons for music & infinity :)

MonsterSwagger:
Both infinity's and music player's icon is not being applied so I included in this list.

Donnnno:
Yeah, I noticed that too & fixed that already ;)

Donnnno:
Opencvmanager's icon is already created by @RuffaWuffa in his last PR so no need to create that one too :)

MonsterSwagger:
Cheers

Rhodney:
The preview for the missing icons:

![preview](https://user-images.githubusercontent.com/5544685/109728078-4c468b80-7b94-11eb-8e42-730443b5ee03.png)

Donnnno:
Lookin' great again!

PR #90, opened 2021-03-02
Supply missing icons from #81

Rhodney:
Color Picker, F-Droid kDI, LibreAV, Shosetsu.


Donnnno:
Great work as always! Thanks :)
Thinking about releasing a new version tomorrow ;)

PR #99, opened 2021-03-13
Use only vector drawables

TBog:
* remove png files
* convert svg to vector drawable using [Svg2VectorAndroid-1.0.2](https://github.com/TBog/Svg2VectorAndroid/blob/jar1.0.2/Svg2VectorAndroid-1.0.2.jar)


Donnnno:
Thank you, I rather handle big changes like this myself now that I know how it works thanks to you
. 
Also, I removed the gradle on purpose because it was fucking with the F-Droid builds, I still have my own just for android studio testing.

PR #105, opened 2021-03-16
Add additional SVGs

Saijin-Naib:
On-going work adding additional SVGs for Arcticons.

### DIFF
--- a/other/a-code.svg
+++ b/other/a-code.svg
@@ -1,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   id="svg22"
+   version="1.1"
+   viewBox="0 0 48 48"
+   height="48"
+   width="48">
+  <metadata
+     id="metadata28">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title></dc:title>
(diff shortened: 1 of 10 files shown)


Donnnno:
Great, but I noticed the artboard size of your target icon is really weird:
   viewBox="0 0 210 297"
   height="297mm"
   width="210mm">

Saijin-Naib:
> Great, but I noticed the artboard size of your target icon is really weird:
> viewBox="0 0 210 297"
> height="297mm"
> width="210mm">

I have no idea what that means 🤣

I'm using the frost template, and scaling the SVGs to fill the boundary of the Square template before I save them out. I did the same thing for the prior icons I submitted.

Is that wrong?

Donnnno:
It should be 48 x48

Saijin-Naib:
> It should be 48 x48

Are what I submitted unacceptable? Or is this just something to correct moving forward?

Donnnno:
Oh no worries, I can fix it myself too. The document size is just really weird. Nothing unacceptable tho.

Donnnno:
This is what I'm talking about: the big rectangle is the artboard size:

image

Saijin-Naib:
This is what I see in my InkScape Document Properties for the main project file I work on, which I attached above (arcticons_square):  
image

That is the expected value of 48x48, correct? I wonder if InkScape adjusts the document size to A4 when I Save As plain SVG instead of the InkScape SVG I'm editing in.

Donnnno:
Yeah, that might be possible, the other icons are fine tho!

Saijin-Naib:
> Yeah, that might be possible, the other icons are fine tho!

My InkScape auto-updates, so maybe this behavior changed from the earlier icons and after the update... Hmm.

I can't seem to get the canvas size to stick after exporting to plain SVG

Saijin-Naib:
I think I'm at a good stopping point for now, if these are acceptable for you.

Donnnno:
Can you provide the lines of code in one xml file, like how turtl does it?

Saijin-Naib:
I've attached the edited XML from TURTL (empty,xml):  
[10_icons.zip](https://github.com/Donnnno/Arcticons/files/6164216/10_icons.zip)

Please let me know if this is what you need.

Donnnno:
Just merged it and checked the icons, be sure to use the 1pt/px line thickness :)

Issue #114, opened 2021-03-26
Broken Icons


ModdedGamers:
I have been using this, and these icons do not display properly. Please fix this!

```
<!-- Aurora Store -->
<item component="ComponentInfo{com.aurora.store/com.aurora.store.view.ui.onboarding.OnboardingActivity}" drawable="aurorastore"/>
<!-- Authenticator -->
<item component="ComponentInfo{me.jmh.authenticatorpro/crc643c2a2b888a14ec9a.MainActivity}" drawable="authenticator"/>
<!-- DuckDuckGo -->
<item component="ComponentInfo{com.duckduckgo.mobile.android/com.duckduckgo.app.launch.LauncherBlack}" drawable="duckduckgo"/>
<!-- SystemUI Tuner -->
<item component="ComponentInfo{com.zacharee1.systemuituner/com.zacharee1.systemuituner.MainActivity}" drawable="systemuituner"/>
<!-- Telegram X -->
<item component="ComponentInfo{org.thunderdog.challegram/org.thunderdog.challegram.MainActivity}" drawable="telegramx"/>
<!-- Weather -->
<item component="ComponentInfo{wangdaye.com.geometricweather/wangdaye.com.geometricweather.main.MainActivity}" drawable="weather"/>

```

dalaeance:
A note on DDG: It won't work if you're using one of DDG's custom icons, which I'm guessing you are based on the `LauncherBlack` bit in the code.

I suppose Donno could add a line for each of DDG's built-in custom icon colours, so people not realizing this don't open issues.

I only know this because the same thing happened to me, but the realization popped into my head right before I included it in an icon request. I probably should've mentioned it then.

ModdedGamers:
Thanks for mentioning that. That's good to know!

Donnnno:
Fixed the remaining broken icons!

PR #115, opened 2021-03-26
Add all DDG alternate icons as filters for DDG app

ModdedGamers:
This adds the purple, green, gold, blue, and black DDG icons to the filters that the app can select for.This adds the purple, green, gold, blue, and black DDG icons to the filters that the app can select for.
This fixes the DDG icon issue in #114


Donnnno:
Thank you!

PR #123, opened 2021-04-07
Add 3 icons

renesat:
I drew Icons for VK, Daylio and NekogramX and update appfilter.xml.


Donnnno:
Looks good! My only suggest would be to remove the box shape around the Dailio and VK icon :)

renesat:
> Looks good! My only suggest would be to remove the box shape around the Dailio and VK icon :)

OK, I fix it.

PR #129, opened 2021-04-16
Fix appfilter.xml

renesat:
In previews pull request I add icons to wrong appfilter. This commit fix it.


Donnnno:
Hi, I think it was my fault. I currently have two appfilter files, so that's a mess. Some launchers had problems detecting the icon pack, so I copied and pasted the appfilter to the assets, I'll make sure it works for the next release. Thanks!

PR #157, opened 2021-05-15
Added Simple Mobile Tools File Manager and Gallery icons

greenpeteuk:
Added Simple Mobile Tools File Manager and Gallery icons and added code to appfilter.xml


Donnnno:
Simple File manager & Gallery already have icons! I think they didn't work because the specific color wasn't added to the appfilter.

greenpeteuk:
You have many gallery app's included in your appfilter file, all seem to point to the same icon. Thus no matter which gallery app you have installed you get the same icon.
If you have (like me) got more than one gallery app, you end up with multiple identical icons in your app drawer, which makes it impossible to identify the app you would like to launch by icon alone, rendering the use of icons usless.
This also appears to be the case with the File managers too (I also have more than one file manager as I can't uninstall or disable the default one).

Donnnno:
Good point! I'll change them along with your pr :)

greenpeteuk:
Thanks! :-)

PR #170, opened 2021-05-27
Add a minetest icon

bydariogamer:
Minetest is a free software game published in Fdroid, so I though it could be cool if this repository had its icon.


Donnnno:
Thank you for your icon! It looks great. 

Can you also use the Turtl app to get the code needed to implement it? 

You can find a more detailed explanation in the contribution doc.

PR #201, opened 2021-06-23
Added new RiseUp VPN activity name

greenpeteuk:
Just updated the RiseUP VPN app' and it reverted back to the old icon.
I checked activity name and it has changed, so I have submitted the code as an addition (not replacement) of the new name, that way it should work for people that don't update ans well as those that do.

Maybe you'd rather have just the new code?


Donnnno:
Thanks, I prefer to keep the old lines in there as well, for older version support :)

PR #217, opened 2021-07-15
Added Tasker Secondary to use tasker icon

ail1020:
I believe I got the syntax correct, but I didn't test it.


Donnnno:
Cool, thanks!

PR #223, opened 2021-07-27
Minor corrections

ail1020:
Blur and Electrum icons were swapped and the RHVoice entry had a typo causing the icon to not show.


Donnnno:
Oof, note to self: Don't try to create an update when you're tired.

Thanks for the heads-up !

PR #224, opened 2021-07-28
Fixes appfilter drawable names of glsbank and ausweisapp

chaptergy:
The drawable names in the `appfilter.xml` were different than what the icons ended up being called. This PR fixes those associations.


Donnnno:
Thanks, I'll merge it right away!

PR #240, opened 2021-08-08
Adds luca and lucia app icons

chaptergy:
luca
lucia


Donnnno:
Hi, I can't open these svg's, did you export them correctly?

chaptergy:
I am able to open the svgs without any issues. I have now saved them in a non-optimized way, maybe then you will be able to open them.

Donnnno:
Ah it works now.

A small suggestion would be to use the lowercase letters because the original app also uses lowercases. But I'll leave that up to you to decide :^)

chaptergy:
You're right, that would fit better. I've changed it, thanks!

Donnnno:
Cool!

Issue #264, opened 2021-08-18
Icon Request


stinglama:
Hi, can you please make icons for the following apps.

- Blackhole
- Join
- Naptime
- SAI
- Truecaller

[IconRequest.zip](https://github.com/Donnnno/Arcticons/files/7008269/IconRequest.zip)

Donnnno:
- [ ]     Blackhole
- [ ]     Join
- [x]     Naptime
- [x]     SAI
- [x]     Truecaller

Donnnno:
Join & Blackhole!

![image](https://user-images.githubusercontent.com/31142286/130428466-2feb5450-5f6a-4292-a170-1014408e0721.png)

Issue #263, opened 2021-08-18
Icon Request Covpass App


tuch96:
Hi, it would be great if you can build up an icon for this new app.

Kind regards!!
[20210818_112438.zip](https://github.com/Donnnno/Arcticons/files/7006147/20210818_112438.zip)

Aman9das:
Pushed to my fork.

PR #266, opened 2021-08-20
#263 #264 add 6 icons

Aman9das:
hopefully usable icons


Donnnno:
Can you rename the icons so you don't have dashes and caps?

Also can you provide an xml with the activity names ?

Aman9das:
Done, and I didn't include infinity in my xml as its an proposed improvement on the existing version

Donnnno:
Hi, thanks!

I've taken a better look at your PR and did some cleanup, here are the changes:

image

A couple of things that I noticed:

- make sure that your line thickness is always 1pt/px
- turn off fill, almost all your icons had a 0% opacity fill.
- use the basic templates as a guide for sizing (don't make them too large)
- and try to use as fewer lines as possible, the covpass icon is a good example for that one. 

Also, I've included the sai & infinity icon as an alt icon, because of the existing ones. You can use them by changing them in your launcher.

Issue #260, opened 2021-08-14
What happened with the icon for GeoGebra Graphing Calculator?


Quantenzitrone:
In the appfilter.xml files in ll. 2282ff is the following:
```
<!-- GeoGebra Graphing Calculator -->
<item component="ComponentInfo{org.geogebra.android/org.geogebra.android.privatelibrary.splashscreen.SplashScreenActivity}" drawable="geogebragraphingcalculator" />
```
The icon for GeoGebra Graphing Calculator doesn't seem to exist in `/icons`.
What happened here?

Donnnno:
Good question! 

I have no idea 🙃 

But I'll check it out later.

Donnnno:
It had the wrong drawable name, it's fixed now!

Quantenzitrone:
Actually geogebra and geogebragraphingcalculator are not the same and also shouldn't have the same icon.
The developer Geogebra publishes 7 apps on the playstore. 5 different mathematical tools and two apps that combine all of them. GeoGebra Classic is the old version and hasn't received updated since 2018 and GeoGebra Calculator Suite is the new version. Geogebra classic and Calculator Suite have kinda the same icon. While all the others (Graphing Calculator, Geometry, 3D Calculator, CAS Calculator and Scientific Calculator) have unique icons.

Donnnno:
Pffff, a lot of different calculators. I only could find the normal one in the icon database, so feel free to draw them ;)

Issue #229, opened 2021-07-31
I need help with this Icon


Quantenzitrone:
![earthmoon](https://user-images.githubusercontent.com/74491719/127748615-eaa38958-1f67-4c0d-a5d7-df5dd50411c8.png)
How could I make this an arcticon-ly?
I tried, but I'm not really happy with it:
![earthmoon](https://user-images.githubusercontent.com/74491719/127749289-9d86c00f-0311-4fc2-8d90-91f6c87a05b0.png) (I had to convert it to png to upload it here)

Donnnno:
Oof, that's a tough one! You're in the right direction tho.

Maybe remove the circles from the moon? 
or just use a simplified globe?

Quantenzitrone:
How about this?
I got the globe and stars from another app in arcticons. I just added the moon, added the sun and moved the stars.
![earthmoonsvg](https://user-images.githubusercontent.com/74491719/129457051-3e328d23-42b9-43b2-8f30-1c7febc68baf.png)

Donnnno:
Looks very good!

Quantenzitrone:
Thanks
I'll open a pull request once I have drawn all the icons I plan to draw.

Donnnno:
Cool!

PR #278, opened 2021-08-27
15 new icons, 1 changed icon & 1 app with an already existing icon

quantenzitrone:
I also updated the appfilter (this time in alphabeti order).

About the changed icon:
In my opinion, the current DB Navigator icon looks more like amogus (very sus) than a train. And since the original icon is now also a "DB" in a square, I made a new icon. If not, I propose the drawable "transportr" also for DB Navigator.

The app with an already existing icon is the one I explained in the issue #260.

The 15 new icons also include the icon from #229.


Quantenzitrone:
Is the CV Project icon also in Version 1.8 because it doesn't work for me?

Quantenzitrone:
oh ok it works in 2.0.5
thanks for making and maintaining this beautiful icon pack

Donnnno:
Ofcourse :^) No problem, thanks for your contributions

PR #282, opened 2021-08-28
Anymex and AnymeUltima

Kaiserdragon2:
Added icons for Anymex and AnymeUltima and added them to the appfilter.xml


Donnnno:
Nice!

But why is has the appfilter so many changes?

Kaiserdragon2:
Oh, I tested the icons and build the app myself and used the sort.sh

Donnnno:
Ah, okay, I'd rather not use the sort function because it removes the dynamic calendar icons from the appfilter list.

Donnnno:
thanks for the fix :)

PR #285, opened 2021-08-30
Last 9 icons of apps i have installed and that don't have an icon (for now)

Quantenzitrone:
Icon drawing is nice. It's fun, relaxing, and you feel like you do something useful.


Donnnno:
Ah yea, I feel you haha.

Donnnno:
Quick question, how important are the letters in the Isotope browser for you? I think it's better without them personally :)

Donnnno:
The same question goes for myvrn ^

PR #301, opened 2021-09-04
Added icons for Craigslist and MuseScore

dgcampbe:
I tried my hand at making some slightly harder icons. The bottom of the "u" for MuseScore probably isn't aligned with the "m", though and might need to be center better (and doesn't have the slight Helvetica-esque "lean" that the original does, but that is probably an improvement from the official icon). The "peace sign" angles probably don't match the original either. Originals are attached for comparison.
craigslist
musescore


Donnnno:
These GitHub previews are really weird :s but when I open them, they look normal.

image

Donnnno:
Aligned and resized some things, other than that, good work! 

image

PR #305, opened 2021-09-04
Added one icon and fixed another, but appfilter.xml needs to checked

dgcampbe:
I am not sure if having two different apps with different icons under the same heading will conflict.
One is a Samsung email ~~bloatware~~ app and the other is a Verizon voicemessage app.
My guess is that my new voicemail icon will need to be renamed and the "drawable" field to the new name, but I don't know if that is the case or not.
Original Icons:
email
voicemail


Donnnno:
Yes, multiple apps within the same entry isn't a problem, I do that all the time. Otherwise, we would have like 6000+ icons :s

Donnnno:
Thanks for the typo fixes :)

Donnnno:
Just checked the appfilter, no spaces dashes or capital letters are allowed in the icon name. Everything else looks fine to me.

PR #315, opened 2021-09-09
bring closer to candybar-sample

Aman9das:
remove the x-hdpi icons as they were absent in candybar sample

Re-add social links..


Aman9das:
@Donnnno  You should definitely check the candy bar sample source...

https://github.com/zixpo/candybar-sample

There you'll see that only nodpi folder has icons.. i added to xxhdpi because icons weren't showing up (i forgot to put appfilter in res/xml)

Donnnno:
There are some different ways to deal with drawables. All the different resolutions are for different devices, so I think it's for the best to leave the original resolutions from mdpi to xxxhdpi in there.

Aman9das:
Okay. . If someone finds that issue, then we can swiftly add the other resolutions back..
btw drawable-nodpi has 256x256 icons... so very high definition..

Donnnno:
Did you reraster them to get it in an even higher res?

Donnnno:
Oh, and can you make different PR's for different additions, that makes it more clear.

Donnnno:
Also there's one icon missing now, the previous count was 2051

Aman9das:
Oh i mean yes, i just pasted the icons from the linux version 256x256.

And yeah i am also wondering what went missing but this is not the finished product so we can port the icons from main again before release...

And i didn't test the themes though...
These are the stuff you should mend:

- In dashboard-configuration, there's an array of social sites.. you can fill them up
- the theme is unfinished, my dad took the computer to office :( midway

Donnnno:
> And yeah i am also wondering what went missing but this is not the finished product so we can port the icons from main again before release...

Your version is behind one commit! there are 2105 icons in total without all the ones I still have to merge for the next release.

Aman9das:
Okay I'm going to edit the raster script for candybar.. and add the missing icons..

Donnnno:
Idk if you have a pc available atm, but I was just planning to do that with my laptop.

PR #353, opened 2021-09-14
Telegram Beta uses the same icon as Telegram

Suburbanno:


Donnnno:
Thanks, I'll merge this.

If you want to edit the appfilter in the future, please use the one in /assets. (I know, it super confusing to have two appfilters)

Aman9das:
I'll add a dummy appfilter.xml, then we can edit that and copy to both the locations in the shell scripts..

Donnnno:
Oh, that's even better!

Issue #385, opened 2021-09-19
Icon request


dalaeance:
[20210919_121545.zip](https://github.com/Donnnno/Arcticons/files/7192474/20210919_121545.zip)

```
<!-- BIWO -->
<item component="ComponentInfo{com.biwo_app/com.biwo_app.MainActivity}" drawable="biwo"/>
<!-- Lime -->
<item component="ComponentInfo{com.limebike/com.limebike.launcher.LauncherActivity}" drawable="lime"/>
```

eninkey:
I'm doing this :)

PR #356, opened 2021-09-14
New activity name for Komoot + some new icons

eninkey:
The Komoot-icon doesn't work on my device. With adding this line, it should fix it, right? (It's from my Turtl export)

PR #398, opened 2021-09-20
2 new icons +  new activity name for music app

eninkey:
Icons for #385
arcticons_nk


joelchrono12:
Hey those are great bud! A lof ot weird commits tho

eninkey:
Thanks. 
Its because of  #356 I deleted the old ones so I dont commit them again. Is there a better way to do it?

PR #401, opened 2021-09-21
TwitchMod uses the same icon as Twitch & Fix Inmage icon

Suburbanno:
TwitchMod uses the same icon as Twitch & Fix Inmage icon


Donnnno:
Thanks!

PR #442, opened 2021-09-25
add a few fun icons

Aman9das:
7 icons #406


Aman9das:
Oops mimo is a mess theres the icon png still behind it :(

Fix it please..

PR #548, opened 2021-10-11
Icons for #365

Kaiserdragon2:


Kaiserdragon2:
I searched the appfilter.xml but hadn't found it but if I am wrong i will delete it.
Have to look into it later.

PR #552, opened 2021-10-12
Update documentation for submitting icon requests

Vinfall:
I'm dumb enough to open a PR in my own folk the day before😞, but now it should be fine.


Donnnno:
Thanks! This is great :)

PR #795, opened 2021-11-13
Make icon requests to-do lists

Aman9das:
Icon reqeststs would be like Github flavored to-do lists this way..


Donnnno:
Ohh what do you exactly mean by that?

Aman9das:
Rn there's no way to track which of the icons in a request have been finished, so it'd be great if the icon requests had a checkbox to track how much of it is finished.

If clash of clans is finished, all of instances can be marked as done..

Donnnno:
Ahhhh, so it adds checkboxes, that's pretty neat :)

PR #828, opened 2021-11-17
added icons

joelchrono12:
Some new icons
showcase


Donnnno:
Looks great as always!
Resized the icons, corrected some line thickness, and removed two doubles: easypaisa & renpho

PR #833, opened 2021-11-20
Some fixes and some new icons

Kaiserdragon2:


Donnnno:
Ah thanks for restoring the android icon, didn't see that I replaced it!

PR #955, opened 2021-12-06
Add six icons for 5 apps

aloxe:
Add six icons for 5 apps


Donnnno:
Cool, great work!
Can you replace the `-` with a `_`?

aloxe:
I hope that was the right file to rename ;-)

Donnnno:
Yes! You saved the app from crashing with that!

Donnnno:
I'm reviewing your icons and niticed that you changed the letter proportion, I changed that back because it doesn't look right.

In general (unless you're the designer for this year's Spotify Wrapped) I'd reccomend to not do that with a font :) 

image

Donnnno:
Anyways thanks for the icons! The rest looks good!!

aloxe:
> In general (unless you're the designer for this year's Spotify Wrapped) I'd recommend not to do that with a font :)

If you mean the font was stretched, this was an not intentional so thanks for correcting it. 

On this one I noticed that in the [letters you provide in templates](https://github.com/Donnnno/Arcticons/blob/main/templates/arcticons_font.svg), the tittle over the "i" twas actually a plain circle without outline. Should it be kept like this or changed into a dot with a 1px outline like all other assets?

Donnnno:
Ah yeah, the dot exception. The only place where you're allowed to use fill is a dot, as long as it's exactly 1,5px wide and tall. It's because dots look better when they're a bit thicker than the lines, some optical sizing stuff.

PR #958, opened 2021-12-08
Nine icons

aloxe:


Donnnno:
Made some changes, to make it more _arcticons_ but overall great work!

PR #961, opened 2021-12-11
Add icons for "Radio FM4" and "Lingvist"

totoroot:
Hey there 👋
I found your project on Mastodon and am loving the icons, so thanks for all the great work to you and the contributors!

Since this is my first contribution to this project and I'm not very experienced with Inkscape, please let me know if I have to fix something.
I read through the contributing guidelines and made sure to use the font as well as snapping to grid wherever possible.
Please double-check the appfilter.xml since I'm not sure, if I found the right activity names for these apps.

Here the original apps as well as the icons I created:

### Lingvist

Looking forward to your feedback and to creating some more soon.


Donnnno:
Hi, thanks for your contribution!  And good to know that you like the icons.

I've taken a look and made some changes, I'll try to explain them as clearly as possible:
- You only have to make the white icons, after you've exported them, you can place them in the `other `folder instead of directly into the `white` & `black` folders. Our scripts handle everything from there, like rasterizing them and making them black.
- The font that you used is not actually used in our icons, we use the raw SVG's for that, and you have to place them manually. You can find that file [here](https://github.com/Donnnno/Arcticons/blob/main/templates/arcticons_font.svg)
- You did the appfilter stuff right!

totoroot:
@Donnnno I'm starting to think that my changes in appfilter.xml were not correct.
Since you've released the new version I installed it from the GitHub release and the Articons app says it's version 3.3.7, but the two added icons do not show up, although they are listed in the Articons app.

How do you get the correct activity names for the application you're designing new icons for?

totoroot:
Screenshot still showing the original app icons

totoroot:
A fix for the wrong activity names for these two apps are part of PR #965

Donnnno:
Ah okay, good to see you've found the solution already :)

PR #963, opened 2021-12-12
Add Airvisual icon

Otemachi:


Donnnno:
Nice work!

Donnnno:
This is great! The sizing is right, so the only thing I had to do is minify the SVG ;^)

Issue #511, opened 2021-10-04
Arcticons Dark Icon Request


Donnnno:
04-10-2021 09:20

App Version : 2.3.6
CandyBar Version : 3.13.0

Bluetooth-Lautstärken Manager
eu.darken.bluemusic/eu.darken.bluemusic.main.ui.MainActivity
https://play.google.com/store/apps/details?id=eu.darken.bluemusic

Cpu Monitor
com.glgjing.stark/com.glgjing.stark.HomeActivity
https://play.google.com/store/apps/details?id=com.glgjing.stark

CPU Spy Reloaded
com.vibhinna.cpuspyfree/com.vibhinna.library.MainActivity
https://f-droid.org/en/packages/com.vibhinna.cpuspyfree/

Device Info HW
ru.andr7e.deviceinfohw/ru.andr7e.deviceinfohw.MainActivity
https://play.google.com/store/apps/details?id=ru.andr7e.deviceinfohw

eWeather HD
com.Elecont.WeatherClock/com.Elecont.WeatherClock.ElecontWeatherClockActivity
https://play.google.com/store/apps/details?id=com.Elecont.WeatherClock

FDE.AI
com.feravolt.fdeai/com.feravolt.fdeai.activities.MainActivity
https://f-droid.org/en/packages/com.feravolt.fdeai/

Fury Survivor: Pixel Z
com.mrxw.android.ltgames/com.mrxw.android.ltgames.UnityPlayerActivity
https://play.google.com/store/apps/details?id=com.mrxw.android.ltgames

Geizhals
at.geizhals.pv/at.geizhals.pv.GeizhalsSplashScreen
https://play.google.com/store/apps/details?id=at.geizhals.pv

GetApps
com.xiaomi.mipicks/com.xiaomi.mipicks.DefaultAlias
https://f-droid.org/en/packages/com.xiaomi.mipicks/

Grüner Pass
at.gv.brz.wallet/at.gv.brz.wallet.MainActivity
https://play.google.com/store/apps/details?id=at.gv.brz.wallet

HANDY PARKEN
at.mobilkom.android.handyparken/at.mobilkom.android.libhandyparken.activities.SplashActivity
https://play.google.com/store/apps/details?id=at.mobilkom.android.handyparken

jetAudio+
com.jetappfactory.jetaudioplus/com.jetappfactory.jetaudio.MusicBrowserActivity
https://play.google.com/store/apps/details?id=com.jetappfactory.jetaudioplus

[...]

PR #965, opened 2021-12-12
Add icon and fix app activities

totoroot:
This PR adds the icon for "Willhaben, the most succesful private marketplace in Austria. Also it fixes the activity names for the two apps I contributed recently as I've since found the correct activity names now with Turtl.

Here is the original app icon as well as the icon I created:


totoroot:
This icon is one of those requested in issue #511

Donnnno:
Looks good! Thanks :^)

Issue #891, opened 2021-11-22
Icon request


Donnnno:
*Info*
1511 0932
App Version : 2.9.1
CandyBar Version : 3.13.0

Airtel
com.myairtelapp/com.myairtelapp.home.views.activities.SplashScreenActivity
https://play.google.com/store/apps/details?id=com.myairtelapp

AuthPass
design.codeux.authpass/design.codeux.authpass.MainActivity
https://play.google.com/store/apps/details?id=design.codeux.authpass

bob World
com.bankofbaroda.mconnect/com.bankofbaroda.mconnect.activities.SplashActivity
https://play.google.com/store/apps/details?id=com.bankofbaroda.mconnect

Canary
io.canarymail.android/io.canarymail.android.dark
https://play.google.com/store/apps/details?id=io.canarymail.android

Tickertape
in.tickertape/in.tickertape.onboarding.SplashScreenActivity
https://play.google.com/store/apps/details?id=in.tickertape

Yono Lite SBI
com.sbi.SBIFreedomPlus/com.sbi.SBIFreedomPlus.sbunifiedcug
https://play.google.com/store/apps/details?id=com.sbi.SBIFreedomPlus 

---
*Zip file here*

[icon_request_15_11_2021_14_01_41.zip](https://github.com/Donnnno/Arcticons/files/7581227/icon_request_15_11_2021_14_01_41.zip)

joelchrono12:
working on this one

joelchrono12:
Done 
![showcase](https://user-images.githubusercontent.com/18448587/146106304-4cdf9e74-0c54-4a87-88f4-72eb73fa9e90.png)

PR #968, opened 2021-12-16
added icons for #891

joelchrono12:
Some icons I did yesterday :P
showcase


Donnnno:
Airtel is a double! 

The rest looks good tho!

PR #969, opened 2021-12-16
Eight icons

aloxe:


Donnnno:
Nice work! 

You could simplify the Lidl plus icon by using the Lidl connect icon and replacing the C with a +

PR #973, opened 2021-12-16
new fairphone 4 app and icon

aloxe:


other/appfilter.xml
```diff
@@ -1558,7 +1558,6 @@
     <item component="ComponentInfo{com.samsung.android.calendar/com.samsung.android.app.calendar.activity.MainActivity}" drawable="calendar_31" />
     <item component="ComponentInfo{org.sufficientlysecure.localcalendar/org.sufficientlysecure.localcalendar.ui.MainActivity}" drawable="calendar_31" />
     <item component="ComponentInfo{org.lineageos.etar/com.android.calendar.AllInOneActivity}" drawable="calendar_31" />
-    <item component="ComponentInfo{com.google.android.calendar/com.android.calendar.AllInOneActivity}" drawable="calendar_31" />
```
aloxe, line 1561: this was a duplicate

Quantenzitrone:
Icon is broken and has a "transform elements" in it, which the [contribution guide](https://github.com/Donnnno/Arcticons/blob/main/CONTRIBUTING.md) asks not have.
The transform elements attribute is also the reason for the R to be so weird.

aloxe:
@Quantenzitrone These are the letters coming directly from the template font so I thought this was allowed in that case. I'll have a look at @Donnnno's fix to figure out.

Donnnno:
Yeah, the lines were acting strange, also the word was too long for one line (so it looked cluttered), so changed that too.

PR #982, opened 2021-12-24
Icons for #451 and some other

Kaiserdragon2:


Donnnno:
Great work! The metro was already present in the pack, but the appfilter is updated now :)

PR #1004, opened 2022-01-24
Added new icons

joelchrono12:
Sorry, a little rusty and with a different distro, but managed to make a couple icons today, might add some more later

image


Donnnno:
Looks great as always tho!

joelchrono12:
Some more icons
image

PR #1026, opened 2022-01-25
Fixed icon names and updated appfilter.xml

Giannis05:


Donnnno:
Hi, I've done something even easier, instead of renaming icons, it's easier to rename the name in the appfilter instead.

Renaming an icon must be done at like 7 different places, so I generally do not recommend that.

Anyways it should be fixed now!

PR #1041, opened 2022-01-27
[appfilter] icon changed: OAndBackupX

yetanother-user:
The icon of [OAndBackupX](https://f-droid.org/en/packages/com.machiav3lli.backup/) is different from the icon of [oandbackup](https://f-droid.org/en/packages/dk.jens.backup/).


Donnnno:
Thanks for the fix!

PR #1053, opened 2022-01-29
fix appfilter for dkb banking (codefactory-version)

csmnn:
The DKB Banking App (Codefactory-Version) has changed its structure. Fixed the appfilter.xml to include the correct activity.


Donnnno:
Cool

PR #1102, opened 2022-03-03
Added 5 new icons

Para-lyzed:
New Icons


Donnnno:
Hi! Congrats on your first PR :) 

I noticed that these lines aren't fully white: 

image

Para-lyzed:
Strange, there must be an issue with my Inkscape install. I went back and fixed 3 files that had off colors, but the colors reverted multiple times before they finally saved properly. Everything should be correct now.

PR #1108, opened 2022-03-07
Fixed a misspelling of my name in all the changelogs

Para-lyzed:
Changed from Para-lized to Para-lyzed


Donnnno:
Oops, thanks for the fix!

Para-lyzed:
@Donnnno It's also misspelled in the release page, but I don't think I can edit that, and I'm not sure if you can either so it's fine if you can't.

Donnnno:
No unfortunately, that's in there till the next update.

PR #1111, opened 2022-03-12
Add one new icon, add an alt icon, fix appfilter

dgcampbe:
If you would prefer the shell scripts to be replaced with python scripts, I could try making a python script to make svg icons that only have letters automatically instead of manually.


Donnnno:
Hey! It's not really necessary for me to have the shell scripts, but if you wanna try making it all better with python, be my guest!

Donnnno:
Btw, you can just use the same hp `hpprinteraio` icon in the appfilter instead of copying.

Donnnno:
is this one ready to review?

PR #1124, opened 2022-03-26
Added Icon for FOSS Warn

ktutsch:
FOSS Warn: https://github.com/nucleus-ffm/foss_warn (de.nucleus.foss_warn)


Donnnno:
Cool! Some feedback:
- you only have to make a white icon, you can place it into the `/other` folder instead of directly in black or white. That get's done automatically when I run a script.
- I''m not a huge fan of outline glyps, try using the exclamation point from our svg font file in `/templates` :)

ktutsch:
I force pushed because of a typo :)

ktutsch:
Force pushed because of a typo :)

PR #1128, opened 2022-03-28
Make the request.txt more Appfilter styled.

Kaiserdragon2:
The convert_request.py is only needed once to convert the current request.txt after that the new email_parser.py can be used to update the request.txt.
The convert_request.py is used exactly like the email_parser.py
grafik

### DIFF
--- a/other/convert_request.py
+++ b/other/convert_request.py
@@ -58,7 +58,7 @@ def removeGreedy(address, element):
 	return element
 
 def parseExisting():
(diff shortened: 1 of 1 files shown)


Donnnno:
Not all heroes capes..

Donnnno:
So the convert one is only needed once?

Kaiserdragon2:
oh you have also declare the appfilter and email path because its only an modifiyed version of the email parser

Donnnno:
Yes! perfect, thank you for making this!

PR #1134, opened 2022-03-31
Added icon for Pleco

Craeckie:
This is my first icon and the stroke seems a bit thick. Is there a way to do that without editing each point?

The apps website is here: https://www.pleco.com/
The activity name is com.pleco.chinesesystem.PlecoDroidMainActivity


Donnnno:
Hi, congrats on your first commit!
Just checked it out, and I noticed a couple of things:

- The size doesn't match, check the template files for the correct size (48x48px)

image

- You're using fill instead of strokes.
- Do not just convert the fill to strokes, it creates very cluttered icons (1)
- If you want to make it fit into Arcticons, try creating lines out of it like example 2 (but with round caps & corners)

image

Donnnno (changes requested):
please follow the contribution instructions.

Donnnno:
closing this, because @EdwinRodger created the icon :)

PR #1139, opened 2022-04-03
Add main activity for SimpleReminder

rYR79435:
The app provides two icons to click on.
One - which was already added - is a shortcut to immediately add a reminder, the other is the actual app.


Donnnno:
thanks!

PR #1143, opened 2022-04-08
Add Icons

tieu1991:


Donnnno:
Made this one a bit more Arcticons. Other than that, good work!

image

PR #1145, opened 2022-04-09
Change Quebec 511

tieu1991:
Change this icon to look more like a road sign, since this is what it is supposed to be.

image


Donnnno:
ah, I see! I'll change it later :)

Donnnno:
fixed this!

Donnnno:
Dunno what happened there! but it;s fixed now, thanks!

PR #1146, opened 2022-04-09
More icons

ktutsch:
A few more icons.


Donnnno:
Hi, Thanks again for your hard work, I changed some small things:

- make sure to center every icon
- check that the line thickness is always 1px 
- try to avoid cluttered icons, I have some examples of what I changed below

(also the doordash d is fully round now.

image
image

PR #1149, opened 2022-04-12
Fix gmail and  google tv and more google fixes + icon(s)

Kaiserdragon2:


Kaiserdragon2:
No Problem  just wondered why gmail was back to "normal" than found some more while checking what happened

Donnnno:
Yeah, I renamed everything to have spaces and thought I did it everywhere too in the appfilter.

Donnnno:
wanted to remove the frame from the Duolingo app too! Didn't;''t work out as intended.

PR #1150, opened 2022-04-13
Made all components fully formed

MrMannWood:
Makes all components pass one of the following regex
- `ComponentInfo\{([^\/]+)\/([^}]+)\}`
- `ComponentInfo\{([^}]+)\}`

This allows stricter parsers to work with these lines instead of discarding them. My guess is that they are current malformed due to copy/paste errors


Donnnno:
This looks really helpfull!

PR #1159, opened 2022-04-17
New icons

ktutsch:
Icons


Donnnno:
What does the F-droid check do?

Donnnno:
> Icons

Lots of em!

ktutsch:
> What does the F-droid check do?

It gets all https://f-droid.org links from the file in the first argument and checks if they return a 404 error or if they exist. Existing links get printed.

Donnnno:
Oh that's pretty cool!

Issue #1163, opened 2022-04-17
Icons got exchanged 


EdwinRodger:
In the latest version (v4.2.0) dark release, app icons for Dimmer and AnimeXstream got exchanged

EdwinRodger:
Closing issue as the PR got merged

PR #1164, opened 2022-04-18
Resolved issue, updated and added some icons

EdwinRodger:
Resolved issue #1163 and
Updated Google Indic Keyboard and Calendar icon component


Donnnno:
Thanks for your fix and the hard work!
(I've applied the renaming to other places too)

PR #1116, opened 2022-03-18
Add files via upload

pawann-2000:
bigbasket logo added

Issue #1122, opened 2022-03-22
Icon request: Buddha Quotes


grrrrr:
*Info*

Icon request for PinDroid

- GitHub: https://github.com/BanDev/Buddha-Quotes
- appfilter.xml: in the attached zip file
- logo: in the attached zip file (and below for reference)

![buddha_quotes](https://user-images.githubusercontent.com/1025506/159528519-a285b458-089c-4241-9875-45618fa75116.png)

---

[buddha_quotes.icon_request_10_03_2022_21_02_44.zip](https://github.com/Donnnno/Arcticons/files/8326184/buddha_quotes.icon_request_10_03_2022_21_02_44.zip)

Donnnno:
Can you use the in app function to request icons? Thanks :)

grrrrr:
Oh, sorry, my bad.  I don't have email on phone so couldn't submit that way.

PR #1175, opened 2022-04-27
Updated appfilter, removed request and added icons

EdwinRodger:
~~No icon .svg icon added~~


EdwinRodger:
Resolved #1116, #1122 and #1134

I would also like to resolve #1171 but I only have app id for Vi because Samsung Members isn't compatible in my phone 😅

dgcampbe:
Is the issue with Samsung Members that you need the activity name, but can't get it without downloading the app? If so, I can try downloading it onto my old Samsung phone. I assume it is only compatible with Samsung phones for some reason.

EdwinRodger:
> Is the issue with Samsung Members that you need the activity name, but can't get it without downloading the app?

Yes

> If so, I can try downloading it onto my old Samsung phone. 

If you can that will be helpful to me and others. Though it is optional as this problem is of other PR, I am just trying to solve it faster

>I assume it is only compatible with Samsung phones for some reason.

I think that's the issue because I tried search it online and didn't get what I wanted
(I have Xiaomi phone)

Donnnno:
These look very good!

PR #1180, opened 2022-05-01
New icons

ktutsch:
I have no idea what I could write here


Donnnno:
> I have no idea what I could write here

icons!

PR #1194, opened 2022-05-09
Added icons for the following apps:

jammo2k5:
AirMusic
aod Notify
GTA III
GTA SA
GTA VC
LunaSea
My Oldboy!
Tap Tap
Vban Receptor
VRTV Free


Donnnno:
Hi, these icons look very good! Thanks a lot :))

PR #1212, opened 2022-06-10
New icons: Jw Library, ViMusic

KristhianX:
I uploaded the icons and updated the appfilter.xml. Icons made with Vector asset creator. Please let me know if there is anything to fix. Thanks in advance!


Donnnno:
Hi, I've taken a look at the files, and fixed the following things:

Your icons were massive (the artboard size should be 38px)

please use the Arcticons font for the letters here
https://github.com/Donnnno/Arcticons/blob/main/templates/arcticons_font.svg

KristhianX:
Hi, thanks for your feedback, I really appreciate it.
Also thanks for your work and this awesome icon pack. I have a few questions:

> Your icons were massive (the artboard size should be 38px)

Does that means I should leave a 5 px margin all around the icon? 

> please use the Arcticons font for the letters here https://github.com/Donnnno/Arcticons/blob/main/templates/arcticons_font.svg

I'm not able to open SVG files in Vector Asset Creator. I downloaded and tried to use the TTF file from https://github.com/Donnnno/Arcticons-Font
but the app will crash every time I try to export to an SVG file. I'll try to use Inkscape in the future, it's not possible for now.

I don't have much experience with GitHub so, what should I do now? Will you change the icons in this pull request or should I close it?

Donnnno:
Ah no worries, I was planning on fixing those things, so no need to do anything :)

KristhianX:
Ohh Ok, thank you! :)

PR #1232, opened 2022-06-30
Add new icons + link existing ones

AbsurdUsername:
Hi 👋

Added icons:
- Graphene Apps
- Intesa (mobile/investo)
- MaterialPods
- Nebbia KWGT
- PingTools
- Tripeaks
- Wolfie for KWGT

I hope `nebbia_kwgt.svg` and `wolfie_for_kwgt.svg` are not too messy/traced, but I wuoldn't know how to draw them otherwise, any tips are apreciated

**Also** it seems like the icons that end with `"/>` instead of  `" />`, with a space, don't apply at all, but I might be wrong and it's something else. Take as example Neo Store, I added the space because right now it's not applying and the componentInfo is correct.


Donnnno:
Hey! Thanks for the contribution, it looks fantastic! (Did you want a change from doing pastel icons 😄 ) 
Both wolfie and nebbia look fine to me, so nothing to worry about there. 

> Also it seems like the icons that end with "/> instead of " />, 

I've added the latest Neo store line a week ago, so it's not in the app yet ;) 

I made some tiny adjustments here and there:

image

AbsurdUsername:
Thanks for the changes, I took some notes :)

> I've added the latest Neo store line a week ago, so it's not in the app yet ;)
Yeah I'm actually an idiot I even looked at the git history, don't mind me

And yes I see you on discord 👀

Donnnno:
Cool! 


> And yes I see you on discord 👀

👀 👀 👀

Donnnno:
Also, just thought of this after taking another look:

image

Replacing lines that are close to each other with just one line makes it a bit clearer, especially when the icons are very tiny.

AbsurdUsername:
Alright got it, thanks 👍

PR #1235, opened 2022-07-02
added several icons

dgcampbe:
Some of these icons could use a bit of touching up because they are off center and have misalignments.


Donnnno:
Thanks!

PR #1244, opened 2022-07-05
powdertoy sentry snapbridge

Calciume:


Donnnno:
Great work
I've made some small adjustments to the icons:

- Connected all the lines for the sentry icon
- turned off all fills that were still present
- shape cleanup for the snapbridge icon (made with circles) 

image

PR #1263, opened 2022-07-14
Add Infinitode 2's icon

andmagdo:


Donnnno:
Hi, the icon itself is fine, but I had to scale and change the color, so start with the template documents the next time please :)

PR #1267, opened 2022-07-15
Fix NekoX App Icon activity name

prirai:
With the introduction of Telegram Premium I suppose they have change the activity from
    `<item component="ComponentInfo{nekox.messenger/org.telegram.ui.LaunchActivity}" drawable="nekogramx" />`
to
   ` <item component="ComponentInfo{nekox.messenger/org.telegram.messenger.DefaultIcon}" drawable="nekogram_x"/>`

The previous versions have the activity name which was previously so I've just added this line below the original one.


Donnnno:
Thanks!

PR #1269, opened 2022-07-15
Add Neo Store-beta icons

prirai:
Note that the previously existing Neo Store icon is different from this beta version and has a different id.
The previous app:
`com.machiav3lli.fdroid`
This beta version which is actually the release candidate and will be continued and the previous discontinued:
`com.machiav3lli.fdroid.neo`


Donnnno:
It looks like the Neo icon we already have, I'd suggest we use the already existing one 

image

prirai:
So great, do I need to do anything on my side?

Donnnno:
Nope, thanks for letting me know:)

PR #1278, opened 2022-07-23
Assign Santander icon to localized app variants.

duese:
Santander logo already exists in this repo but it needs to be assigned to localized variants of the Santander banking apps.


duese:
Created one PR subsuming multiple PRs: #1281

PR #1291, opened 2022-07-28
Fixed appfilter Chess PGN Master

ojppe:


Donnnno:
hi, the appfilter is already correct like this, so closing this one

PR #1302, opened 2022-08-10
NekogramX drawable name correction

prirai:
In continuation to a previous commit #1267 where I had a pull request for Nekogram X in which we had to change component and drawable, I had the drawable set as `nekogram_x` which will clearly not match the icon kept in the icons directory. Therefore, made the required change.


Donnnno:
Hi, thanks, you can use the already existing neo-store icon tho, no need to add a new one :)

prirai:
Updated as required. GitHub is really clunky.

Donnnno:
already fixed in another commit! but thanks :)

PR #1305, opened 2022-08-13
Four logos added

Kerioneesh:
Logos added for 4 apps I use on my phone: Huawei Mirror, Meditation, Yettel (previously called MyTelenor) and MÁV.
Previously sent two icon requests via email, both sent on 4th August, 12:40 CET. Figured I can design them myself. I used the SVGOMG tool to remove the unnecessary metadata.


Donnnno:
did some small tweaks, but good work!

PR #1310, opened 2022-08-20
Add three icons, modify one and various activity fixes

totoroot:
Hey @Donnnno :wave: 

While taking a look at all the changes in Inkscape 1.2 (loving the update so far :heart_eyes: ), I worked on some icons for apps that do not yet have pretty representations in this wonderful icon set.

Here you can see the original icons that I based my designs on:

  

I also found the Sennheiser to not be very recognisable. So I changed it for the associated app "Smart Control" and added an association with "CapTune", which is another app by Sennheiser, that uses the brand logo as the icon.

The new design looks more like the brand's logo:

I hope I put everything in the right directory this time around.

Once again thanks for continuing to work on this :bouquet:


Donnnno:
Hi! Thanks for the great work again :-)

PR #1312, opened 2022-08-24
Added icons for MovieLab and Seal

Lavin-tom:
Added icons for MovieLab and Seal, updated xml file of Vi


Donnnno:
image

Which one is the movielab icon? you left some others in there :)

Donnnno:
great!

Issue #1289, opened 2022-07-28
Icon request JEFIT


basings:
````
	<!-- JEFIT -->
	<item component="ComponentInfo{je.fit/je.fit.SplashScreenActivity}" drawable="jefit"/>
````

![jefit](https://user-images.githubusercontent.com/92867705/181495945-12e1ef30-49ca-454d-8045-453f971dda80.png)

best I can make is

![jefit](https://user-images.githubusercontent.com/92867705/181498080-533a5315-588b-463a-87b6-0ea3b7c486ca.svg)

Issue #1288, opened 2022-07-28
Icon request qdslrdashboard


basings:
The icon for qdslrdashboard is really complex and maybe out of my skillset

![qdslrdashboard](https://user-images.githubusercontent.com/92867705/181489155-3726b2c9-37bb-4794-8ce9-bf8b2b6c5694.png)

````
	<!-- qDslrDashboard -->
	<item component="ComponentInfo{info.qdd/info.qdd.DslrDashboard}" drawable="qdslrdashboard"/>
````

unless following is an acceptable solution?

![qdslrdashboard](https://user-images.githubusercontent.com/92867705/181495666-a6a9e13c-9a61-4197-8ede-4aef7d2db12c.svg)

or

![qdslrdashboard2](https://user-images.githubusercontent.com/92867705/181495692-7bf65e3e-102c-4740-8ca6-d4ceed5a1f56.svg)

Issue #1287, opened 2022-07-28
Icon request Clear Outside


basings:
Creating the icon is just too much for my current skills

![clear_outside](https://user-images.githubusercontent.com/92867705/181488584-e31210ed-903a-4fd2-a560-6b42d7f71020.png)

````

	<!-- Clear Outside -->
	<item component="ComponentInfo{com.firstlightoptics.clearoutside/com.firstlightoptics.clearoutside.MainActivity}" drawable="clear_outside"/>
````

basings:
![clearoutside](https://user-images.githubusercontent.com/92867705/181500638-ecb10d3b-39e2-4adf-a887-ead487612e71.svg)

not too bad, is it?

PR #1314, opened 2022-08-27
added icons

basings:
- clear outside
- hotukdeals
- jefit
- qdsldashboard

updated
- mull


basings:
I wasn't sure if I should delete the mull icon in /icons/black and /icons/white ?

basings:
solves #1287 #1288 #1289

basings:
added 

- streetcomplete_expert_edition

original icon is in a different color than the icon for streetcomplete so no changes for arcticons.

Donnnno:
can you use the already existing streetcomplete or streetcomplete_alt icon instead?

basings:
sure! I wanted to do that first but then I thought if they change the icon in the future, it's only an icon change and not an appfilter change. Hence, I copied the icon. I'll change it later

basings:
done

Donnnno:
Thanks, I get the point, we can always update it in the future if needed :)

Donnnno:
> I wasn't sure if I should delete the mull icon in /icons/black and /icons/white ?

No, it gets automatically replaced!

PR #1317, opened 2022-08-28
add some icons, link some more

GitGitro:


Donnnno:
nice!

Donnnno:
Just noticed that we already have a f1 logo, so we can use that one (you can also modify that one to have something like 'tv' beneath it, or a tv around it, idk).

Also, one last thing for the 711 logo, you can remove and connect it to each other to make it more minimal:

Schermafbeelding 2022-09-01 om 09 21 25

Issue #811, opened 2021-11-13
Icon request


Donnnno:
*Info*
0911 1148
App Version : 2.8.6
CandyBar Version : 3.13.0

BAWAG PSK
com.bawagpsk.bawagpsk/com.bawagpsk.bawagpsk.ui.splash.SplashActivity
https://play.google.com/store/apps/details?id=com.bawagpsk.bawagpsk

BusBahnBim
de.hafas.android.stv/de.hafas.main.HafasApp
https://play.google.com/store/apps/details?id=de.hafas.android.stv

digi4school
com.vsb.digi4school/com.vsb.digi4school.MainActivity
https://play.google.com/store/apps/details?id=com.vsb.digi4school

Digitales Amt
at.gv.oe.app/at.gv.oe.app.MainActivity
https://play.google.com/store/apps/details?id=at.gv.oe.app

Drohnen-Info
at.oeamtc.drohnen/at.oeamtc.drohnen.MainActivity
https://play.google.com/store/apps/details?id=at.oeamtc.drohnen

Drone Space
aero.unifly.austrocontrol.launchpad/crc64b96ef428d1bce9e2.MainActivity
https://play.google.com/store/apps/details?id=aero.unifly.austrocontrol.launchpad 

---
*Zip file here*

[icon_request_09_11_2021_11_48_17.zip](https://github.com/Donnnno/Arcticons/files/7532117/icon_request_09_11_2021_11_48_17.zip)

PR #1320, opened 2022-08-30
Add two icons and a couple icon associations in appfilter.xml

totoroot:
Hey @Donnnno :wave: 

I'm still exploring Inkscape 1.2 in my free time and once again I worked on some icons for apps that do not yet have pretty representations in this wonderful icon set.

Here you can see the original icons that I based my designs on:

 

They were both part of issue #811.

Someone added a beautifully crafted icon for the "George" app with the latest update to the icon pack, but only added the association with the localized Romanian app. So I went ahead and added the remaining localized versions that I found on the Google Play Store.

Also for the Owncloud/Nextcloud "Notes" app I added a changed icon association, because it got lost with a recent update to the app.

Thanks for your work! :clap:


Donnnno:
Oh that's great! Cool :)

PR #1322, opened 2022-08-31
add german bank app

Mrandyrew95:


Donnnno:
Hey! Can you rename it to nameofbank_banking instead? Every banking app calls it just banking and that causes confusion for the pack ;)

Mrandyrew95:
@Donnnno Resolved, thanks :)

PR #1328, opened 2022-09-02
Update Syphon Matrix client icon

GrantMoyer:
Syphon icon was still based on the old icosahedron icon. This change updates the icon to be more in line with Syphon's current official icon.

See https://github.com/syphon-org/syphon/commit/28e05bc73adb3de9f16a283c85aea853ae9f34d2

Rendered:

| Existing | Updated |
|:----:|:----:|
| syphon exisiting | syphon updated |


Donnnno:
Nice!

PR #1334, opened 2022-09-14
Updated xml of Popcorn time

Lavin-tom:


Donnnno:
it should be 'popcorntime' like the entry above. Can you fix that?

PR #1341, opened 2022-09-21
add new icons

GitGitro:


Donnnno:
great set of icons again, thanks for the hard work! 

just did some mini sizing tweaks :)

PR #1346, opened 2022-09-25
coopvoce, docusign, everli_shopper, houzz, ufficiopostale, wekiwi

plasmid-zz:


Donnnno:
Nice work!

Can you use our svg font for the ufficiopostale icon?

also I think that the coopvoce icon works without the text too :) it's more of an icon that way

plasmid-zz:
Thank you. Here the modified svgs. I'd like also to modify everli_shopper as a single line vector, but I think it's too much difficult.

coopvoce
ufficiopostale

Donnnno:
Great

PR #1366, opened 2022-10-29
update M.A.L.P appfilter

dgcampea:
ref: https://gitlab.com/gateship-one/malp/-/commit/0ebbcf5cddc904fef4e59f0b879bd942ec09daae


Donnnno:
Nice, you only need to update the one in the other folder tho :)

PR #1371, opened 2022-11-08
Updated appfilter

Lavin-tom:
Updated xml of Organic Maps


Donnnno:
Thanks, updated it in #1372 !

PR #1372, opened 2022-11-08
appfilter: fixed 2 errors and updated 2 apps to new componentinfos

Quantenzitrone:


Quantenzitrone:
oh whoops #1371 already updated organicmaps

Donnnno:
Thanks for adding the entries!

PR #1373, opened 2022-11-08
add tap wizard 2 icon

aspwil:
this is for the app tap wizard 2 https://play.google.com/store/apps/details?id=com.topcog.tapwizard2&hl=en_US&gl=US

i talked with the app dev and he approved of using this icon for the app in the icon pack.

it is based off of the staff seen in the icon for the steam game version. (as the normal app icon is pretty complex, and would be difficult to turn into a good wire icon)

the svg file still has metadata from inkscape in it, as i don't know what parts i should clear, so i just left it all. as it says "Usually Inkscape will add a lot of metadata that will get cleaned up by the mantainers" in the inkscape contribution guide.

let me know if i messed anything up, or it does not conform to the standards laid out in the contributing guide document.


Donnnno:
Great, it's all good like this, thanks!

PR #1379, opened 2022-11-16
Added 11 icons

rmpj5:
Added Icons for Apps from F-Droid and Google Play:
- 1822direkt Banking
- EmiliosDict
- Expense Register
- GPS Test Plus
- Kleine Wettervorschau Deutschland
- Mopria Print Service
- PasswdSafe
- SecureGo+ Renault Bank direkt
- ADAC Trips
- Verkehr.NRW

Updated filter + updated icon to match the new icon from the app
- ADAC Spritpreise

Overview


Donnnno:
Hi! Can you remove all capitals from the icon names? 

image

It makes the app charsh :s

rmpj5:
Done. I've also removed the + sign, just in case.

Donnnno:
yes thank you!

Donnnno:
Hi!

I'm doing some small tweaks rn, I'm not a huge fan of text hen the icon should speak for itself... is the ADAC text really necessary here?

image

rmpj5:
I see what you mean. I had added the text because it is part of the icon delivered with the app. This makes it easier for users to recognise their apps. But the text could probably be removed from the upper (spritpreise) without doing harm.
Without additional text, the lower icon (trips) would be very hard to distinguish from other icons already in Arcticons. The icons for Map and Mapsgeobookmarks would look very similar, Baidu Map is also visually very close. So, I'd keep the text but the final decision is yours. I've attached the original icon for your reference. 
Trips_icon

Donnnno:
Yes, thanks for your reply, good one!

Donnnno:
did some small tweaks, fewer lines here and there

image

PR #1381, opened 2022-11-19
Fixed the display of the sound recorder icon on some devices

Artem13327:
the original icon without fix is a white circle with a red dot in the middle


Donnnno:
no need to add it in any other folders than the `/other` folder the next time :)

THanks!

PR #1382, opened 2022-11-19
added icons for Banking4, CDisplayEx

jakejoh:


Donnnno:
resized the icons, other than that, great work!

PR #1387, opened 2022-11-21
Add 4 icons

rmpj5:
Add icons for Amplerbikes, Denkzettel, HE Networktools and RNZ Newspaper. 

I've tried to find a balance between your style and the icons delivered with the apps to facilitate recognizability. 

Overview textlogos


Donnnno:
Hi! Thanks for your work, we only accept line icons, if we add other shapes, it won't be as consistent as it is now..

So, can you modify these icons? Thanks!

image

rmpj5:
@Donnnno Done. 
Overview textlogos update

Donnnno:
Ah, nice work! Thanks for the quick fix :-)

PR #1389, opened 2022-11-25
Add icon metadata for all versions of shuttle

GrantMoyer:
Add icon metadata to reuse the icon used for Shuttle+ for both Shuttle and Shuttle 2


Donnnno:
Thanks for adding the entries!

PR #1393, opened 2022-12-03
Map more app versions on appfilter

joserebelo:
- Debug version of Catima
- chaldeaprjkt's GameSpace
- IP Webcam Pro
- Termux:Tasker
- Yubico Authenticator changed the activity


Donnnno:
Thanks for updating the entries!

PR #1400, opened 2022-12-14
Hotfix

Kaiserdragon2:
This should fix the missing appfilter entries


Donnnno:
Thanks for fixing my mistakes ! :)

PR #1419, opened 2023-01-01
Update appfilter.xml

fix-icons:
Currently OLX India doesn't have any icon, though Olx.ro's can work very well for it as well.

(Their original icons have minimal differences. OLX India have just 'India' at the bottom.)

Hoping this is the right way to do it.


Donnnno:
Great, thank you!

PR #1422, opened 2023-01-01
3 new apps

artsuaga:


Donnnno:
Hi! Nice work :-)

I think the icons would work too without text in them, like this:

image

What do you think>?

artsuaga:
Sure, I see them well

PR #1425, opened 2023-01-02
scripts for new release rewrite

Kaiserdragon2:
This needs some work.
Do you use the raster or the raster-nosort ? 
it is possible to implement both and it will be determined which one to use through an argument passed to the script.


Donnnno:
Oh, didn't expect to see something like this! That's awesome :) 

I use them both, raster for releases and nosort for quick tests.

Donnnno:
Just a quick thought, I'm using a newdrawables.sh script too, but those aren't added to the drawable.xml files automatically, would it be possible to add it to that too?

Kaiserdragon2:
Do you mean that the newdrawable.xml is moved to the asset folder?
At the moment the newdrawable.sh is called from the raster.sh should this be handled the same ?

Donnnno:
Oh I mean the `newcatagory.sh` script, it generates a list called `newdrawables.xml` and places it in the other folder like this:

```
<?xml version="1.0" encoding="utf-8"?>
<resources>
	<version>1</version>
	<category title="New" />
	<item drawable="abnto" />
	<item drawable="act_fibernet" />
	<item drawable="acv" />
	<item drawable="alpimaps" />
	<item drawable="always_visible_volume_button" />
</resources>
```

I was wondering if this script could be integrated in the process by placing them in the _new category_ in the `drawables.xml` file that's generated

Kaiserdragon2:
Sure that's possible 
That would be the function that the merge_new_drwables.py script does at the moment

Kaiserdragon2:
Ready to use for info how to use just enter "python preparerelease.py -h"

Donnnno:
Coooool! Gonna test it soon :)

Donnnno:
Hmmm what am I doing wrong here? 

image

Kaiserdragon2:
Oh sorry the description for the second argument should be without the /main only 
/App/src

Donnnno:
Wow! It works like a charm, and it's superfast! 

The only thing I noticed is that all the new entries in the drawable are double now, idk if it's easy to fix otherwise my check duplicate plugin will do the trick too ;)

Kaiserdragon2:
i just did run the script myself and i don't have duplicate entries. It could be that because the script run into in error the first newdrawables.xml was added to the second one. I added a line to delete the old file before creating a new one

Donnnno:
Still finding duplicates, very strange

image

You can see them in the in-app icon count too

Donnnno:
Also found some weird appfilter things going on:

image

- Enters are removed (not really an issue)
- Names starting with a lowercase letter are put on the bottom now
- Names with special characters are really weird now :s

Kaiserdragon2:
> Still finding duplicates, very strange
> 
> image
> 
> You can see them in the in-app icon count too

Ahh I know what's happening here the new icons are added to the new category and the respective alphabetical category. 
The In-app count just counts the entries in the drawable.xml

I will add something that the new icons are only in the new category.

Donnnno:
> Ahh I know what's happening here the new icons are added to the new category and the respective alphabetical category.
> The In-app count just counts the entries in the drawable.xml

So the drawable list needs to be generated before the new images are moved to the folder where it counts from and after that, the newdrawables needs to be put in there?

Donnnno:
> I will add a line that checks if the drawable is in the new category and only add them if it is not in the new category.

Cool, thank you so much!!!!

> i couldn't get it working with them being added while sorting i didn't implemented it.

Fine by me, it still works so that's the important part

> In the "old appfilter the lowercase characters are also at the bottom

I thought they were sorted alphabetically at least

Donnnno:
Gonna merge it because the important things work! 

I still enjoy the all-in-one solution and it's so fast too.

Some future things to think about would be:
- Spaces between entries
- Correctly sorting everything alphabetically: lowercase words are put at the bottom now, the same goes for characters with special glyphs like Ä

PR #1426, opened 2023-01-02
Update appfilter.xml

fix-icons:
Icon for Bromite doesn't show in recents screen (unlike other apps). It seems the activities differ for the launcher and while in use.

Found the right activity class using adb and appended to the current entry. If this doesn't work, I'll make another PR to revert the changes.


Donnnno:
Sounds good!

PR #1431, opened 2023-01-10
Added a few icons

gbucar:


Donnnno:
Hi, thanks for the icons, I've made some small adjustments to make it fit more into Arcticons:

image

Mostly some resizing, but also changing the font to Arcticons one :)

PR #1432, opened 2023-01-11
Two existing logo added for two apps

Kerioneesh:
Hi there!
I have added two existing logos for two apps. Only changed the appfilter.xml file.


aloxe:
Happy to see my bank icon can be reused. Raiffeisen is present in many countries, I hope they'll be more.

Donnnno:
Nice!

PR #1439, opened 2023-01-26
added 4 icons

joelchrono12:
Added some icons! Because why not? :P
showcase


Donnnno:
Nice!

Donnnno:
Did the thing here: 
image

PR #1446, opened 2023-02-08
Added 1 icon

RCaled:
I added 1 icon from a game. I hope it's okay. It's my first time doing something like this.


Donnnno:
Nice!

PR #1459, opened 2023-02-26
New & Updated Icons, tried to fix #1458

### LINKED ISSUE #1458: Icons not applying
Hi,
for the following apps icons don't apply on my phone, despite existing in Arcticons:
(information from AppRequest)
TK app:
`<item component="ComponentInfo{de.tk.tkapp/de.tk.tkapp.start.StartActivity}" drawable="tk"/>`
MAPS.ME:
`<item component="ComponentInfo{com.mapswithme.maps.pro/com.mapswithme.maps.MainActivity}" drawable="mapsme"/>`

Best regards

KR4U7H3R:
Hey this is my first pull request and I hope I did everything correctly.

Firstly I want to say that I am not sure what to do with old entries in appfilter.xml. As some icons were not applying on the current app version, but when downgrading to older versions, the icon pack did work (Cinelog F-droid). So I did not delete anything in appfilter.xml.
Some designs might need some additional work. I designed mostly the size of the original icons, I guess some could could be enlarged to fit better with the icon set.

- Filmstarts: icon and added to appfilter.xml

- yetCalc: icon and added to appfilter.xml

- Updater:  icon and added to appfilter.xml (not sure about the name of this one. I am using xiaomi.eu ROM, but it seems to be just called like this)

- Cinelog: updated appfilter.xml 

- Dicio: updated icon and appfilter.xml

- Chefkoch: updated appfilter.xml

- jtx Board: updated appfilter.xml

- Maps.me: updated appfilter.xml (Tried to fix #1458)

- TK: updated appfilter.xml (Tried to fix #1458)

- Sportschau: icon and added to appfilter.xml. 2 Versions: the original icon looks like Option 2, but I guess with the Text it does not work.

- Upgrado: icon and added to appfilter.xml

- MyETraining: icon and added to appfilter.xml

- Clementine-Remote: icon and added to appfilter.xml

- mLauncher: icon and added to appfilter.xml. In the request.txt I found this:

<item component="ComponentInfo{app.mlauncher/app.mlauncher.MainActivity}" drawable="mlauncher" />

my Phone gave me different values (I got the Fdroid version). So I did just added my phone's path. Possible that the other has to be added as well.
And I know that the icon does not really follow the guidelines... As I am not that talented in Inkscape I did not know how to improve it. Maybe Just a capital "M"?

Then I started to work on ARD Audiothek:

ARD Audiothek and ARD Mediathek (which is already in Arcticions) have the same icon, just the color is switched. How do you treat such cases?
My ideas:

1. use the same icon as Mediathek and add "Audiothek" (not centered)
2. change both icons to the same style with writing (I guess this does not really fit with the icon pack) (centered)
3. add some symbol for Audio? (Just an example)
4. just use the same icon? (difficult: because of similar name they appear right next to each others)

Thanks for all the work, it is an amazing icon pack, and by adding/fixing these, my phone will be 100% covered. First icon pack I found, which achieved this. :+1: 

Regards


skedastically:
Hi, saw your icons. I think we can only use Arcticon's fonts, so I suggest less text.

**ARD apps**: what about just the number 1 logo? Then add the letters A or M on bottom-right corner to separate the two apps. Or even better, symbols of headphones/video.

**Sportschau** without text seems ok.

I also made yetCalc in another PR. Lemme know what you think

KR4U7H3R:
Thanks for the feedback.

ARD apps: that's an idea but there is also tagesschau, which already has this icon. Most (but one) of the icons i have on my phone are more or less recreations of the original icon, but I could also live with an icon as you described it.

yetcalc: ah yes, that is unfortunate. Looks good as well, slightly bigger, which could be useful. We can go ahead with yours, I do not care if mine gets included or not.

Donnnno:
Hi, nice work! I've introduced some changes to your icons to make them more _Arcticons:_

image


Also, please use the SVG alphabet in the template folder next time, otf, ttf fonts don't work with Arcticons.

PR #1460, opened 2023-03-01
Added 3 icons

RCaled:
Added icons for 2 games and for an app. Sorry, I accidentally closed my old pull request after deleting my fork.


Donnnno:
Nice work! Translating anime icons to lines is always a struggle, but you've found a good way to do it.

I've just changed the kuro icon to be a bit more minimal:

image

PR #1470, opened 2023-03-12
Fix a couple of oddities in appfilter.xml

bsvka:
I don't know why several entries used regular parenthesis instead of the curly braces, but it tripped up my script. The drawable `_2_player_games.svg` exists in the `/icon/black`, but was misnamed here. I'm assuming this'll fix the app icon (it should be broken right now).

The changes are entirely untested, as my use-case for the file is different from the regular icon pack.


Donnnno:
thanks for the fixes!

PR #1472, opened 2023-03-12
updated xml files

Lavin-tom:


Donnnno:
Samsung calendar already has a separate icon here:

image

PR #1485, opened 2023-03-26
Hardening the release script

Kaiserdragon2:
In some cases stroke and fill definitions are written slightly different  now these cases should also work


Donnnno:
Nice!

PR #1487, opened 2023-03-27
Added another version of the Twint app (prepaid)

a-zurcher:
Added missing prepaid version of Twint app in appfilter.xml

### DIFF
--- a/other/appfilter.xml
+++ b/other/appfilter.xml
@@ -18370,4 +18370,45 @@
 
   <!-- ﾆｺﾆｺ動画 -->
   <item component="ComponentInfo{jp.nicovideo.android/jp.nicovideo.android.StartupActivity}" drawable="niconico"/>
+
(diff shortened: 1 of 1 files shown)


Donnnno:
Hi, adding a bunch of appfilter strings without icons doesn't work.

Can so add the twint lines (if it has an icon) to the already existing twint lines? 

This part is also unnecessary: 
```<!-- ADDED BY a-zurcher - END -->```

a-zurcher:
> Hi, adding a bunch of appfilter strings without icons doesn't work.
> 
> Can so add the twint lines (if it has an icon) to the already existing twint lines?
> 
> This part is also unnecessary: `<!-- ADDED BY a-zurcher - END -->`

Hello, you are totally right, I used the wrong commit for this pull request, I'll close this one and open another with the correct commit. Sorry about that !

PR #1493, opened 2023-03-29
Added another version of the Twint app (prepaid) - second try

a-zurcher:
Hello, I've added a single line in appfilter.xml for the prepaid Twint app (which already has an icon).


Donnnno:
Thanks!
I'll merge it right away :-)

PR #1496, opened 2023-04-03
Add 3 icons

20brs:
I've added the SVG files for Blinkist, Cyrus 2 and Tranzy, along with their entries in other/appfilter.xml


Donnnno:
Nice!

Donnnno:
For some reason, the appfilter is acting strange! So adding this manually in another commit, thanks for your icons!

PR #1516, opened 2023-04-29
Add additional package name for F-Droid version

Hell01World:
Dungeon Crawl Stone Soup 
https://f-droid.org/en/packages/org.develz.crawl/
See line 5048


Donnnno:
Cool!
Can you update the main appfilter the next time directly? Thanks :-)

PR #1517, opened 2023-04-29
Added 2 icons

ojppe:
[Wurdian](https://play.google.com/store/apps/details?id=com.QuixoticApps.Wurdian)
[Chess PGN Master Pro Key](https://play.google.com/store/apps/details?id=com.kalab.pgnviewerpro)


Donnnno:
Hi! nice icons :-)

For some reason, there was an appfilter file in the master folder, I've deleted that one because we use the one in the /other folder :)

PR #1519, opened 2023-04-30
Add 1 icon

dimassasin1:
svg width="48" height="48" version="1.1" viewBox="0 0 48 48"
I'll probably upload one so that the rest do not have to be redone, in case this one is not ok.

Tabletki.ua
com.tabletkiua.tabletki/com.tabletkiua.tabletki.main.SplashScreen
https://play.google.com/store/apps/details?id=com.tabletkiua.tabletki


Donnnno:
Nice work!

PR #1522, opened 2023-05-01
Added 24 icons

Axonym:
I hope these are in line with contributing guidelines.
Please let me know if they aren't and what I could do to improve my work.


Donnnno:
Hi! 

Thanks for your contribution, great icons!

Some feedback points, nothing big:
- Line-width should always be 1px (Inkscape is probably the reason for why they weren't)
- **Kideapp:** Circles should always follow the same sizes 43px (illustrator), 44px (Inkscape).
- **One Second Diary:** This one has many details, I would simplify in this scenario.
- **Pivo:** if you make a word, letters should be of equal base height.
- **Tuudo, Seabank:** Personal preference: small letters are harder to read on a small screen, so I would use our font there.
- Removed Balance because we already have one for that.
- I tend to put loose words in boxes, to make it more icon-y.

Anyways, I've fixed these issues, and overall fantastic work!

PR #1526, opened 2023-05-02
Keepass2Android Offline: Update appfilter.xml

yetanother-user:


Donnnno:
Hi, be sure to check the drawable name in that line of code, because it's different from the line above it.

Donnnno:
Thanks!

PR #1535, opened 2023-05-11
14 new icons

Stephane-74:
I hope it works this time ! :-)


kaanelloed:
You need to add the activity of your new icons in the appfilter.xml file, not replace all the content with them.

Donnnno:
Thanks a lot for your icons, they look amazing!

Just a couple of things why it took longer to add them:

- Please don't overwrite the appfilter with only your lines of code, just add them somewhere in between.
- Sizing could use a bit more attention, take a look in the template folder or at the contributing file for more info about that 
- Make sure that your branch is always as up-to-date as possible with the main branch (here)

PR #1537, opened 2023-05-13
Retry two dots + add easycontacts and tubelabs

aloxe:
Two dots app is not displaying the right icon despite having the right component name. I am clearing the culter above to see if this fixes it.


Donnnno:
Nice work!

PR #1539, opened 2023-05-14
Rename Blacklist To Deny-List

Saijin-Naib:
This is more in-line with the stated ethics and reflects recent understanding around inclusive verbiage in OSS projects


Donnnno:
Commented in the other PR on this, but thanks a lot!

PR #1540, opened 2023-05-14
Add/Enable Files Shortcut

Saijin-Naib:
This icon has a shortcut overlay to distinguish it from the Files app/icon


Donnnno:
Deny-list is definitely a better word choice, thanks! 

Can you add the svg to the /other folder instead?

Saijin-Naib:
Gah, sorry for the extra commit in this branch. Let me see about dropping it and force-pushing. I didn't want to mix the commits

PR #1546, opened 2023-05-18
Add two more Paperless apps

bsvka:
They are both frontends of Paperless and use a nigh-on unaltered version of the official logo.


Donnnno:
Can't view your changes because you've overwritten the appfilter probably, can you rebase and add the lines again?

PR #1547, opened 2023-05-20
Add two more paperless apps

bsvka:
They are both frontends of Paperless and use a nigh-on unaltered version of the official logo.


Donnnno:
Thanks!
Added them in a fresh PR because of some merge errors

PR #1551, opened 2023-05-20
Add M&T Bank

Saijin-Naib:
Who knows how many folks use this

### DIFF
--- a/other/appfilter.xml
+++ b/other/appfilter.xml
@@ -9679,6 +9679,9 @@
   <!-- Länsförsäkringar -->
   <item component="ComponentInfo{se.lf.mobile.android/se.lf.mobile.android.module.outside.ProxyActivity}" drawable="lnsfrskringar"/>
 
+  <!-- M&T Bank -->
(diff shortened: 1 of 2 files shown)


Donnnno:
Hey, It seems like you used the .TTF font, but that one isn't compatible with the icons, you should use the .svg and place each letter manually.

Because they are only words, I'll do it for these, but in the future it would be great if you can use the SVG font :-)

https://github.com/Donnnno/Arcticons/blob/main/templates/arcticons_font.svg

Saijin-Naib:
Hmm, okay, thanks.

How do I keep the alignment and kerning consistent if I'm placing the letters manually?

Donnnno:
_with feeling_

Saijin-Naib:
This is going to be a disaster :rofl:

Donnnno:
> _with feeling_

To give a bit more context, kerning is a tedious hard thing to get right, and at the end it's all about feeling

https://method.ac/type/

Donnnno:
Also, the icons:

image

PR #1560, opened 2023-05-28
optimized pngs

Donnnno:


Donnnno:
just compared the latest release with optimized PNGs..... 1.7 MB difference 🙃

PR #1565, opened 2023-06-05
Some Icons and appfilter fixes

Kaiserdragon2:


Donnnno:
Nice work! The peridot one is cute~!

PR #1567, opened 2023-06-09
2 icons + regional

Redbox3070:
added 2 icons and updated appfilter to match some icons to regional variants


Donnnno:
Hi! Thanks for the icons, some feedback:

the icons right now have some double lines, I think it could be more minimal, like this (first icon):

image


Also, something's going wrong with the appfilter, you might wanna redo that and add the lines of code with a text editor like VS Code

Donnnno:
Nice work!

Redbox3070:
> Nice work!
thanks. sorry if i doo too many commits but i'm not familiar with github. 
trenord icon was already available but it was not showing so i updated the appfilter, payback.svg icon was already available in a variant so mine is not necessary

Donnnno:
I understand! It looks good now :-) 
Thanks for the fixes

Donnnno:
Going through your icons and noticing that Genius scan already has an icon 

image

PR #1572, opened 2023-06-12
Update libretube entry in appfilter.xml

Dunexus:
Required since https://github.com/libre-tube/LibreTube/releases/tag/v0.15.1 and https://github.com/libre-tube/LibreTube/pull/3915/files


Donnnno:
Nice!

PR #1574, opened 2023-06-12
Material You

Kaiserdragon2:
add XML Converter to prepare release script for Material You version
The icons are exported to a new folder "/you/res/drawable-anydpi-v31"
There are some things needed to be done for this to work as a new flavor.
Preview ☟ 
https://github.com/Donnnno/Arcticons/assets/8929967/64b11bd1-a4f6-4c32-8562-c4582c33e20d


Donnnno:
woowww! insane!!

I was thinking about maybe making the stroke slightly thicker like 1.2 for example, would that be possible with the converter?

Kaiserdragon2:
sure it's just one number change

Kaiserdragon2:
I think it should be without the background because it's like you said not the Arcticons style and also with the launchers ability to set background you are free to set one but it wouldn't be possible to get rid of one. Another possibility would be an additional Version with background.

Donnnno:
Yeah haha, let's leave it to the user to decide, We can probably inform them about it in the app description, gonna think about that

Kaiserdragon2:
This can be archived with the same script it just has to be modified a bit

Kaiserdragon2:
Debug Apk for testing
[APK-Debug.zip](https://github.com/Donnnno/Arcticons/files/11725968/APK-Debug.zip)

Kaiserdragon2:
did you try pip3 install svgpathtools

Donnnno:
okay after a frustrating hour of trying to figure it out,  python -m pip install did the trick.

Also python was installed in the inkscape folder 🤨

Donnnno:
I did use the script tho, and batch converted everything, but the lines are still white?

Donnnno:
Downloaded the debug apk! Great work, also these slightly thicker lines work quite nice!

Donnnno:
Idk why, but android renders it pretty blurry

Donnnno:
Oh I was wondering if a dark/light theme switch is possible

Kaiserdragon2:
You mean an automatic switch between light and dark based on the system inside the app?

Kaiserdragon2:
> but

Ah I misunderstood you. The script i send was for the dark and light arcticons so it uses xml icons instead of png. In this Pr the the dark and light are still pngs and the you version uses xml icons. It might be possible to change the dark/light pngs also to xml but i haven't tested android version compatibleness.
The script also isn't an universal converter because it searches for some data in the svg while other things are predetermined.

Donnnno:
> You mean an automatic switch between light and dark based on the system inside the app?

Inside and outside the app.

I downloaded pix material you light/dark, they're able to switch colors based on your system preference. Would be pretty cool to have that too, I guess.

Kaiserdragon2:
New Debug version with slight color change between Dark and Light Mode
[APK-Debug.zip](https://github.com/Donnnno/Arcticons/files/11735641/APK-Debug.zip)

Donnnno:
Nice nice!

Donnnno:
I've tested it for like a day now, but are doubting about the use of a background. It goes against the original Arcticons designs, but we still have those ofc, they aren't going anywhere. I do think that it's better for Material You to have a background, thin-lined icons can be pretty hard to see with different backgrounds, especially when the color also looks like it. Also, from a popularity perspective, I think people like Material You with backgrounds more than it is right now. 

But I'm curious what others think about this.

Kaiserdragon2:
> One problem I have is that candybar seems to ignore the inset in the icon drawable which results in cutoff icons inside the app. if the option
> 
> > <bool name="includes_adaptive_icons">true
> > is set and another shape the system default is chosen everything works fine
> 
> image

Fixed it

Kaiserdragon2:
Now with support from api level 26 then only a dark light mode icon change is implemented
[device-2023-06-14-200930.webm](https://github.com/Donnnno/Arcticons/assets/8929967/fbfefb09-ddf1-41e2-a6f5-eb29e15f2166)
[APK-Debug.zip](https://github.com/Donnnno/Arcticons/files/11749359/APK-Debug.zip)

Donnnno:
What do you think, are we ready to merge this? 👀

Kaiserdragon2:
With the widgets fixed, i think this is ready to go. Just some minor improvement mentioned below should be taken care of.

I will do an additional PR for the icons.
The main image should definitely been fixed i will try to look into it.

I leave the color choice for the icon to you. Maybe ask on matrix for some opinion. 

And we should add a monochrome version for the main icon, i think there isn't one yet.

In the description and maybe also inside the app we should mention that not every launcher refreshes the icons on a theme change. Probably list some that do the trick.

Donnnno:
Yes, I'll take a look at those!

Kaiserdragon2:
I have tried some other xml graphics and all seem to be displayed in a low resolution maybe it is a problem with the candybar dashboard?

Donnnno:
Candybar limits XMLs to 256 because icons don't usually go above that number. If we wanted to change that, all numbers would have to be specified. So, I don't think it's worth doing that work. Gonna create a PNG instead :)

Donnnno:
What do you think of this one? Delta contributor @creepertron95 came up with the idea :) 

image

PR #1575, opened 2023-06-12
Add multiple icons

tieu1991:


Donnnno:
Nice, those skiff apps hare hard to distinguish from each other tho haha

PR #1578, opened 2023-06-14
update splitser in appfilter.xml

WJehee:
Hey!
Just a single update to appfilter.xml
I just added the new line, I'm not sure if that's how it's supposed to be or should I remove the old entries?


Donnnno:
No that's it, thank you!

PR #1579, opened 2023-06-14
5 new icons

saintpou:
Added 5 icons, 4 of which that are requested icons.
I'm pretty new to GitHub and also vector design, so point out anything I've messed up or things I should've done better.


Donnnno:
Hey! I can't merge right now, because of the appfilter file? Did you edit it with notepad or something? If you should add it with a code editor like VS Code

saintpou:
The new file uses LF line ending now but the problem remains, what seems to be the issue?

Donnnno:
I think because of all the changes in the past week! Maybe it's the easiest if you rebase your fork and add it again?

saintpou:
Roger that

PR #1581, opened 2023-06-14
Added Icon for OYO

tellmeY18:


Donnnno:
Hey! I can't merge right now, because of the appfilter file? Did you edit it with notepad or something? If you should add it with a code editor like VS Code

PR #1588, opened 2023-06-16
Added Icon LPU Touch

knsrinath:


Donnnno:
Thanks for adding the appfilter, but make sure to edit it with a powerful text editor like VS Code, standard notepads don't go that well. Can't see the changes right now :/

PR #1589, opened 2023-06-16
Add icon for LPU Touch

knsrinath:


Donnnno:
Hey, I noticed you closed this one. I wanted to comment that you need to edit the appfilter with an editor like VS Code or Notepad++ instead of a regular notepad. Right now there are too many changes, as you can see.

Issue #1580, opened 2023-06-14
[You] Bugged icons reporting


Donnnno:
This issue is for reporting bugged icons in the new Arcticons You flavor

Donnnno:
- Anylist
- Concentricclock

Donnnno:
- nebula

Donnnno:
- jitten
- oscilloscope
- pulseaudiortpreciever
- tata neu
- translate you
- treecard
- volumio

I think that I've found the most prominent errors now

PR #1592, opened 2023-06-18
Some Optimization 

Kaiserdragon2:
This should reduce cases with invisible icons or invisible parts like those in  #1580
Fixed some existing icons


Donnnno:
Nice!

PR #1595, opened 2023-06-20
Add a few icons

Aman9das:
The UNO icon uses clips please check if it is alright to work with..


Donnnno:
Hey! Better not use clipping masks, I removed them and minified the icon :)

PR #1608, opened 2023-07-04
Adding 4 app icons

Salmonmon:
Broccoli
Bible in one year
Feudal tactics
Get eduroam

(first pull request, please let me know if any bits are not quite right)


Donnnno:
Nice work!
Just did some resizing :-)

PR #1617, opened 2023-07-12
Fix missing icons

wwwwwwari:
This pull request mostly fixes icons not showing up in the following ways:
1. **Rename some icon files** - for cases where it is obvious that the icon files are just a typo/weirdly named, 5fe19d2029cf15ec48491ecc6e57ba723d549166 renames the files in `src\dark`, `src\light`, `src\you`, `icons\white` and `icons\black`.
    * Hyperrail (`hyperraril` -> `hyperrail`)
    * Mi Fitness (`mi_fittness` -> `mi_fitness`)
    * Pepper (`pepper_1` -> `pepper`)
2. **Add new icons** - for cases where I couldn't find the existing icon files that the appfilter is trying to use at all, 7ebe5507c7f54d55a0b1f0528ecb9ede7e107ac1 adds new icons.
3. **Change the drawable value in some appfilter entries** - for apps that I could find the existing .svg files they're trying to use, but the `drawable=` value is wrong, 6f4519cf4de167a2591e0f1a372f41c53026067a simply updates the `drawable=` to match.
    * Note: This also adds 3 other variants of Lien Quan Mobile that I found.


wwwwwwari:
About the the first item, I found the file names also being mentioned in:
- app/src/main/res/xml/drawable.xml
- app/src/main/res/xml/iconpack.xml
- app/src/main/res/values/iconpack.xml
- app/src/main/assets/drawable.xml

I think I'll update those too just to be safe 🤔

Donnnno:
Thanks for all the renaming fixes! 
ps. you don't have to edit the iconpack.xml and drawable.xml because they get generated each time with a new release :-)

Donnnno:
Heyo, the icons are looking fantastic again!

Two things I found while checking the code:

`stroke:rgba(255,255,255,.9998)`  I found this in previous commits too, must have been some kind of rounding error because Illustrator doesn't detect this too.

`stroke-dasharray:none;stroke-opacity:1;stop-color:#000"`
Also found these elements, removed them because they seem unnecessary.

PR #1570, opened 2023-06-11
Add 21 new icons and update appfilter

wwwwwwari:
Hello, this pull request includes the following changes:

1. Add 21 new icons (mostly Thai apps) and new appfilter entries
2. Add new variants of 7-Eleven, Blokada, Gallery, PDF Viewer, Shopee, and Tetris to use the existing icons 
3. Updated Gallery Go in appfilter.xml to use google_gallery_go.svg (previously pointed to gallerygo.svg which does not exist)

Some icons may look 'complicated' here (I tried) - particularly glo_lottery and we_mahidol. Let me know if I need to adjust anything.

PR #1628, opened 2023-07-22
Add support for rgb and rgba values in preparerelease script

wwwwwwari:
This commit adds rgba and rgb value format support in fills and strokes for `preparerelease.py`.

### Background
In #1570, I submitted icons for `tungngern` and `true_iservice` apps. The `strokes` in the SVGs contained non-white values in rgba format but the script did not catch the error:
* [tungngern.svg](https://github.com/wwwwwwari/Arcticons/blob/f8aad34329aff025a1408bd771bc70516d477e74/other/tungngern.svg): `stroke:rgba(255,255,255,.9994)`
* [true_iservice.svg](https://github.com/wwwwwwari/Arcticons/blob/f8aad34329aff025a1408bd771bc70516d477e74/other/true_iservice.svg): `stroke:rgba(255,255,255,.9995)`

The script then let these icons passed through other functions and produced incorrect output files:
1. They're written to `icons/black` and `icons/white` without any changes in the `strokes` values: [1](https://github.com/Donnnno/Arcticons/blob/d4c21ffc11cf756514ff9b265097e346e8923ae8/icons/black/tungngern.svg), [2](https://github.com/Donnnno/Arcticons/blob/d4c21ffc11cf756514ff9b265097e346e8923ae8/icons/black/true_iservice.svg), [3](https://github.com/Donnnno/Arcticons/blob/d4c21ffc11cf756514ff9b265097e346e8923ae8/icons/white/tungngern.svg), [4](https://github.com/Donnnno/Arcticons/blob/d4c21ffc11cf756514ff9b265097e346e8923ae8/icons/white/true_iservice.svg)
2. They're also written to `app/src/dark` and `app/src/light` incorrectly - the PNGs were simply empty: [1](https://github.com/Donnnno/Arcticons/blob/d4c21ffc11cf756514ff9b265097e346e8923ae8/app/src/light/res/drawable-nodpi/tungngern.png), [2](https://github.com/Donnnno/Arcticons/blob/d4c21ffc11cf756514ff9b265097e346e8923ae8/app/src/light/res/drawable-nodpi/true_iservice.png), [3](https://github.com/Donnnno/Arcticons/blob/d4c21ffc11cf756514ff9b265097e346e8923ae8/app/src/dark/res/drawable-nodpi/tungngern.png), [4](https://github.com/Donnnno/Arcticons/blob/d4c21ffc11cf756514ff9b265097e346e8923ae8/app/src/dark/res/drawable-nodpi/true_iservice.png)

While these were manually caught and partially fixed later*, I think it'd be nice if the script can catch the issue automatically.
(*note: their files in `icons/` are still non-white and in `app/src/dark/` are still empty as of now; I can submit corrected files for them later)

### Issue Summary
1. The script does not support `rgba` values in `stroke` and `fill` correctly. If the values are non-white, the script does not error out. If the values are white, the script cannot replace them with `#fff` and `#000` like it would with other formats it supports (e.g. `white`, `#FFFFFF`, `#ffffff`)
2. Likewise, the script also does not support `rgb` values, which are also a valid format for `stroke` and `field` in SVGs.
3. Additionally, the script also does not support decimal points in `stroke-opacity` and `fill-opacity` and will error out as non-white even if the value is, for example, 0.00 or 1.000 (i.e. only integer 0 or 1 is supported).

### Changes
1. At lines 42-45, add `|rgb\(255,255,255\)|rgba\(255,255,255,1\.?\d*\)` to the regex for `ORIGINAL_STROKE`, `ORIGINAL_STROKE_ALT`, `ORIGINAL_FILL` and `ORIGINAL_FILL_ALT`. This adds `rgb(255,255,255)` and `rgba(255,255,255,1*)` as alternative color values to be replaced by `#000` and `#fff` in `svg_colors`. The 4th value in `rgba` may be 1, 1.0, 1.00, 1.000, and so on.
2. In `find_non_white_svgs`:
    1. At lines 453-454, change the regex of `stroke_colors` and `fill_colors` to include rgb values (rgba will be handled by new separate regex's)
    2. At lines 458 and 463, add `rgb(255,255,255)` to the arrays of possible values to check the colors against.
    3. At lines 468 and 473, at another condition to check if the opacity is really 0 and 1, but with one or more decimal points consisting of only zeroes (i.e. 0.0, 1.0, 1.000000 as opacity should be considered valid even if the values are not integers)
    4. Add `stroke_rgbas`, `fill_rgbas` and their respective `for` loops to specifically handle rgba values 

Please let me know if there are any questions or concerns. I know this script is very important so it's okay if this takes some time to be reviewed.

### DIFF
--- a/other/scripts/preparerelease.py
+++ b/other/scripts/preparerelease.py
@@ -39,10 +39,10 @@ VALUE_PATH = APP_SRC_DIR + "/main/res/values"
 #Export Sizes of the icons
 SIZES = [256]
 #Define original color
-ORIGINAL_STROKE = r"stroke\s*:\s*(#FFFFFF|#ffffff|#fff|white)"
-ORIGINAL_STROKE_ALT = r"stroke\s*=\"\s*(#FFFFFF|#ffffff|#fff|white)\""
-ORIGINAL_FILL = r"fill\s*:\s*(#FFFFFF|#ffffff|#fff|white)"
-ORIGINAL_FILL_ALT = r"fill\s*=\"\s*(#FFFFFF|#ffffff|#fff|white)\""
+ORIGINAL_STROKE = r"stroke\s*:\s*(#FFFFFF|#ffffff|#fff|white|rgb\(255,255,255\)|rgba\(255,255,255,1\.?\d*\))"
+ORIGINAL_STROKE_ALT = r"stroke\s*=\"\s*(#FFFFFF|#ffffff|#fff|white|rgb\(255,255,255\)|rgba\(255,255,255,1\.?\d*\))\""
+ORIGINAL_FILL = r"fill\s*:\s*(#FFFFFF|#ffffff|#fff|white|rgb\(255,255,255\)|rgba\(255,255,255,1\.?\d*\))"
+ORIGINAL_FILL_ALT = r"fill\s*=\"\s*(#FFFFFF|#ffffff|#fff|white|rgb\(255,255,255\)|rgba\(255,255,255,1\.?\d*\))\""
 #Define Replace Colors
 REPLACE_STROKE_WHITE = "stroke:#fff"
 REPLACE_STROKE_WHITE_ALT = '''stroke="#fff"'''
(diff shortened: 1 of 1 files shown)


Donnnno:
Really useful, thank you!

PR #1632, opened 2023-07-28
One app had a wrong drawable. Fixed it.

PixelRuzt:
The "drawable" for NFC Tasks was incorrect and I have corrected it in this commit.


Kaiserdragon2:
You should check the appfilter.xml it seems like there went something wrong again.

PixelRuzt:
Okay, I'll check it. for now I'll just delete this pull request then.

PR #1633, opened 2023-07-28
Add icon: ChatGPT (OpenAI)

NicoWeio:
> ChatGPT
> com.openai.chatgpt/com.openai.chatgpt.MainActivity
> https://play.google.com/store/apps/details?id=com.openai.chatgpt 

It's definitely not perfect, but I hope it helps nonetheless. :)

PR #1636, opened 2023-08-01
added 17 icons

dioxias:
Hi! I made 17 icons, and two apps have two icons because I'm unsure which version would fit better. if they both do, great! however if only one does, that's also okay :)

Icons I made:

- characterai
- chatGPT
- claw
- color palette
- flare
- jersey mikes (two versions)
- little caesars
- mrepo
- myCOMPASS
- nasa (two versions, one based on the WORM logo they use)
- plain app
- realtor.com
- scp anywhere
- simple mp
- trulia


dioxias:
Added cromite as well, so it's up to 18 icons

Donnnno:
Nice, thank you!

NicoWeio:
For future reference: The ChatGPT icon was merged in #1633.

PR #1641, opened 2023-08-04
redditsync → lemmysync

NicoWeio:
There is now _Sync for Lemmy_, which “replaces” _Sync for Reddit_.
The icon remains the same, and I have done the following:
- renamed the icon (and adjusted references to it) to _lemmysync_, the new app
- added said new app to the `appfilter.xml`

Please let me know if you'd like to have this done differently.


Donnnno:
3 more icons needed renaming :-) but thanks!

PR #1650, opened 2023-08-15
Adding three new icons (BambuHandy, VVSMobil, RegioRadStuttgart)

mmaeusezahl:
I added three new icons that I was missing. Used them on my smartphone locally in the past week and I'm happy.
The third one still seems too detailled but I tried to stay somewhat true to the original.
Also I tried to "hygenize" the SVGs as much as possible.

all

If you have any ideas for improvement I'm happy to update them prior to merging.

Thanks for the cool icon pack anyway :)


Donnnno:
Great work!

PR #1651, opened 2023-08-15
add icon for blood pressure monitor

NobodyForNothing:
app-id: `com.derdilla.bloodPressureApp`

[app](https://github.com/NobodyForNothing/blood-pressure-monitor-fl)

If I forgot something, please say so.


Donnnno:
Thanks!

PR #1653, opened 2023-08-16
adding 2 icons

Crazy-Unicooorn:
Since my first pull request got accepted, I made 2 new icons!
- adventure_lab
- subscriptions

I think I made a mistake making my svgs 150*150px, and anyway I think my svg code looks very different than the other I can see in this project, but not sure if anything should get fixed by hand or not.


Donnnno:
Nice work again!

The artboard size should be 48px indeed :-)

PR #1661, opened 2023-08-20
AnonAddy -> Addy.io

ThatOneCalculator:
AnonAddy is now addy.io and has a new logo

Made with Inkscape, ran through [SVGOMG](https://jakearchibald.github.io/svgomg/)


Donnnno:
made it a bit more minimal:

image

PR #1675, opened 2023-08-27
add a bad icon, add entries for freedesktop, fix vendetta

dgcampbe:
Split 1 icon into 2, add an icon that needs some touching up, add icons to freedesktop.
The icon that I tried to make is supposed to look like this, and needs someone who knows what they are doing to fix or something:
sente_online_go


Donnnno:
Nice!

PR #1685, opened 2023-09-03
Adding new icons for Pro Darts Counter and SSB apps

mmaeusezahl:
Icons are already minified and sanitized as good as possible.
Appfilter tested in debug build (Preview below is taken from screenshot on my phone).

icon_preview


Donnnno:
Looks good!

PR #1688, opened 2023-09-04
freedesktop-theme: skip icons in mapping file but not in icons folder

autinerd:
freedesktop-theme: add creation of symbolic files by converting with inkscape
freedesktop-theme: add a few mappings


Donnnno:
Idk what it does but I'll merge it anyway 🙃

PR #1693, opened 2023-09-09
freedesktop-theme: new mapping system, icons are now only generated for symbolic and scalable.

autinerd:


Donnnno:
Wow nice work!!!

Donnnno:
Hmmm I think these changes may have broke some KDE things because a lot of icons aren't working anymore

PR #1698, opened 2023-09-16
Added icon for ShareNow app and fixed one component name

mmaeusezahl:
- The icon is also shared with the FreeNow app so I added this as well. (The FreeNow app actually has a different background color so I didn't know how to accomodate for that. I think it is okay if they share the same icon.)
- The SSBFlex app I contributed [here](https://github.com/Donnnno/Arcticons/pull/1685) changed its main activity name

Preview of the new icon:
icon_preview


Donnnno:
Nice!

PR #1725, opened 2023-10-07
Add Network Analyzer Pro to appfilter

pgaskin:
Use the same icon as the free version since it's just a slightly different color.


Donnnno:
Cool

PR #1729, opened 2023-10-08
Fix nfc_reader icon

petlyh:
The nfc_reader icon was rasterized weirdly by Inkscape. The problem did not occur when using a different rasterizer (resvg). Not sure what caused it, so I just removed and recreated the part that was having issues, which seems to have fixed it.


Donnnno:
noticed it too, thanks for the fix!

PR #1731, opened 2023-10-09
Added an icon for the "My Tello" app

MikeTheGreat:
I'm not sure that I've got rounded line caps (all of my paths are continuous circuits, if that helps).  Is there a way to bulk-fix this in Inkscape?

Any feedback is appreciated!


Donnnno:
Hey!
Idk how to do that in Inkscape, but it should be possible.

As for the icon itself, it has a weird gradient? 
Also, we prefer or own font within the icons, so I've changed that :-) 

image

image

MikeTheGreat:
This all looks great!  Thank you for the fixes (and I'll use the standard font for future icons :)  )

No idea where the gradient came from - I was aiming for white lines, but I'll be more careful in the future.

Thanks!

PR #1747, opened 2023-10-15
Stroke auto 1 or 0

Kaiserdragon2:
If stroke with is near 1 or 0 it will be automatically applied to 1 or zero instead of printing it to the console

### DIFF
--- a/other/scripts/preparerelease.py
+++ b/other/scripts/preparerelease.py
@@ -475,12 +475,25 @@ def remove_svg(dir:str):
 
 # Check Icons
 def checkSVG(dir: str):
+
+    def replace_stroke(match):
+        strokestr = match.group("strokestr")
+        stroke_width = float(match.group("number"))
+        if stroke_width > 0.9 and stroke_width < 1.2:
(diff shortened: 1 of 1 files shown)


Donnnno:
Does this have any impact on SVG's without the stroke_width property (because the default amount is 1)

wwwwwwari:
While we're at it, can we have an auto color rounding to `#ffffff` as well? InkScape loves to mess with the color 😂

Kaiserdragon2:
if there is no stroke defined (so it is the default 1) nothing changes only something like 0.999 will get 1

Kaiserdragon2:
> While we're at it, can we have an auto color rounding to `#ffffff` as well (e.g. from `#fafafa` `#fbfbfb` and so on)? InkScape loves to mess with the color 😂

at the moment the script checks it, but it needs to be manually edited and i don't know if every case is catched with this because there are many methods to define a color and to automatically change something that's near white to white for every of this cases isn't easy to do without fucking up in some cases

PR #1751, opened 2023-10-19
OutOfMilk, SuperCycle

MikeTheGreat:
I found the 'rounded corners' and 'rounded caps' options in Inkscape so the paths should end correctly.


Donnnno:
Hi, thank you for your contributions!

Just from looking at it in the browser, I noticed two things:
- capital letters in your names: Can you remove those and add spaces between the words using `_`, otherwise the app crashes.
- the cycle icon has black lines!

PR #1780, opened 2023-11-06
Pocketbook icon

onuronsekiz:
Corrected wrong app icon, deleted duplicate_1


Donnnno:
was in the middle of preparing the update, so I've changed it there! Thanks for the heads up :)

onuronsekiz:
wow, so an update is really coming,
I can finally delete tons of apps that I would never use.
Btw, would f-droid update come late again?

Donnnno:
Yes, everything is ready, but want to make sure everything is good!
If you have android studio, you should be able to test the new version already. :-) 

F-Droid usually takes a couple of days, yea.

Donnnno:
550 new icons and 8068 in total btw, fantastic work!

PR #1782, opened 2023-11-08
Update appfilter.xml

onuronsekiz:
2 apps that already has icon but doesn't have appfilter entries.

Maçkolik
Digiturk Play Yurtdışı

### DIFF
--- a/other/appfilter.xml
+++ b/other/appfilter.xml
@@ -6518,6 +6518,9 @@
 	<!-- Digiturk Play -->
 	<item component="ComponentInfo{com.digiturkwebtv.andstb/com.digiturkwebtv.stb.MainActivity}" drawable="digiturk_play"/>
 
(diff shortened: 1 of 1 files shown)


other/appfilter.xml
```diff
@@ -6518,6 +6518,9 @@
 	<!-- Digiturk Play -->
 	<item component="ComponentInfo{com.digiturkwebtv.andstb/com.digiturkwebtv.stb.MainActivity}" drawable="digiturk_play"/>
 
+	<!-- Digiturk Play Yurtdışı-->
```
Donnnno, line 6521: Nice, no need to add an extra name then, right?

onuronsekiz: Yeap, icons and files are same.
They only need appliter input.

Donnnno: image

I actually meant doing like this haha ^

PR #1787, opened 2023-11-10
Fix typo

supertsy5:
I'm not sure whether such typo is normal for a non-German user...


Donnnno:
The d in donno stands for dyslexia

Donnnno:
looks good like this, thanks for the fix!

PR #1791, opened 2023-11-13
Fix Missing and extended Coverage

TotallyAvailable:
The Watchface Pro entries might never come into play, the App doesn't yet exist.

Would've added Cromite WV (doesn't and might never exist) and Thorium as well while I was at it. Sadly (and purely personal) with Chrome 120 moving into Stable in 3 weeks, using 117 + the severe delay in the 64bit Version release hasn't really sold me on giving it a go yet as many people seem to love it though, despite the concerns surrounding it.

Why that would stop me from just grabbing the required strings manually ? Might anyway later, unless you plan on pushing the renaming-release like right now.


Donnnno:
Fantastic!

PR #1795, opened 2023-11-13
fix #1794: errors in dropbox print and signal icons

### LINKED ISSUE #1794: icons print, dropbox and signal are faulty
print.svg has this weird edge
image

signal.svg has sharp edges except for one
image

dropbox.svg has `stroke-linecap:round;stroke-linejoin:round` missing resulting in sharp edges

quantenzitrone:
closes #1794


quantenzitrone:
Does this work like this? The next release will replace these in icons/{white,black} right?

Donnnno:
Gonna take a look at a different approach for the signal icons, the rounded strokes come too close to each other right now.

quantenzitrone:
i reworked the signal icon, i think i didn't quite get the rounding at the bottom right, but inkscapes dashed paths makes the dashing pretty easy...

Donnnno:
I swear dashes are gonna drive me insane one day. 

Wasn't quite happy with the dashes, so tweaked those to stay true to the original shape.

PR #1806, opened 2023-11-16
update one appfilter entry

dgcampbe:


Donnnno:
Didn't notice it haha

PR #1813, opened 2023-11-18
Update appfilter.xml

TotallyAvailable:


Donnnno:
No need to remove those spaces, our preparerelease script handles that every release :)

PR #1819, opened 2023-11-20
Add Soundbound Icon

Subroutine7901:
Just adds an icon for the app Soundbound (in.shabinder.soundbound).


Donnnno:
Hi, you also need to add some lines of code to the appfilter.xml file, otherwise it won't get applied. 
You can gather it with icon request: https://play.google.com/store/apps/details?id=de.kaiserdragon.iconrequest

Subroutine7901:
Ok so here's the code, I'm just putting this here so I don't forget or lose it:
`<appfilter>

<item component="ComponentInfo{in.shabinder.soundbound/in.shabinder.soundbound.MainActivity}" drawable="soundbound"/>
</appfilter>`
Otherwise, a quick rundown of what I need to do with the modified appfilter.xml, and the required edits to it would be appreciated.

Subroutine7901:
Ok, I think that's it.

Donnnno:
Fantastic, thank you, and @TotallyAvailable for helping out :)

PR #1822, opened 2023-11-20
rename requests.txt to requests.xml

kitters-kat:
edit: changed to draft since the file contents might need to be changed or the pr be discarded since the extension isn't proper xml

### DIFF
--- a/other/scripts/email_parser.py
+++ b/other/scripts/email_parser.py
@@ -1,14 +1,14 @@
 """
 Script Usage:
-python (or python3) email_parser.py ./path/to/emlFolder ./path/to/appfilter.xml (./path/to/requests.txt)
+python (or python3) email_parser.py ./path/to/emlFolder ./path/to/appfilter.xml (./path/to/requests.xml)
 
 Arguments
 0: Path to folder containing .eml files of requests
 1: Path to existing appfilter.xml to recognize potentially updatable appfilters
-3 (optional): existing requests.txt file to augment with new info
+3 (optional): existing requests.xml file to augment with new info
 
 Output
-If only two arguments are given the script will generate 'requests.txt' and 'updatable.txt'.
+If only two arguments are given the script will generate 'requests.xml' and 'updatable.txt'.
 If the third argument is given the file will be overwritten with the updated info.
 """
 
@@ -95,7 +95,7 @@ def findZip(message):
             zip_data = part.get_payload(decode=True)
             zip_file = io.BytesIO(zip_data)
             return  zip_file# Only save the first zip attachment
-    return None                
+    return None
 
 def greedy(message):
     # Get the sender's email address
@@ -114,7 +114,7 @@ def showGreedy():
     for sender in email_count:
         if email_count[sender] > requestlimit:
             print('---- We have a greedy one: '+ str(email_count[sender]) + ' Requests ' + sender)
-    
+
 def parseEmail():
     # Create a new root element for the combined xml
     combined_root = ET.Element('combined')
@@ -143,12 +143,12 @@ def parseEmail():
                 sender = message['From']
                 no_zip[sender] = mail
 
-                        
+
 
 def requests(child,msg):
     global data
     if child.get('component') == None:
-        #clear data for new entrty 
+        #clear data for new entrty
         data = {}
         child_string = ET.tostring(child, encoding='utf-8').decode()
         name_match = re.search(name_pattern, child_string)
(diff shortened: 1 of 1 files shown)


petlyh:
I'm not sure this is a good idea considering that the file is not valid XML. File extensions should generally reflect the file's content accurately. Having a `.xml` extension also causes editors to attempt to interpret it as XML, which looks like this for me in VS Code:



I think the extension should be kept as `.txt` unless the format of the requests file is changed to be valid XML.

dgcampbe:
Would reworking the file's contents to match a proper xml format be a good way to resolve that problem?

petlyh:
> Would reworking the file's contents to match a proper xml format be a good way to resolve that problem?

Not entirely sure. Changing the requests file to an XML-based format would potentially make it easier for the script to parse, but it would also make the requests file more verbose (maybe not ideal, considering that it's already ~4.5 MB).

I also think it could be confusing to have request info represented as XML mixed together with the actual `ComponentInfo` XML that should be copied into `appfilter.xml`. Here's a structure mock-up that illustrates the problem, though maybe a different structure could improve it:

```xml
<!-- OnStream -->
<request>
  <item component="ComponentInfo{com.maertsno.m/com.maertsno.m.ui.main.MainActivity}" drawable="onstream"/>
  <gplay url="https://play.google.com/store/apps/details?id=com.maertsno.m"/>
  <fdroid url="https://f-droid.org/en/packages/com.maertsno.m/"/>
  <times-requested>14</times-requested>
  <last-request-time>1700051822.0</last-request-time>
</request>
```

I personally think the current text-based requests format is fine since the script seemingly has no problems with it, but it's up to @Donnnno.

Also, may I ask what the motivation for this PR was?

dgcampbe:
I mentioned something related in the Arcticons Linux matrix room and hastily came up with the idea to rename the Arcticons Android requests file without even actually checking if it was proper xml. Derp on my part lol. It might still be an interesting idea to have the requests file be in a format that can be more easily operated on by programs.

petlyh:
> I mentioned something related in the Arcticons Linux matrix room and hastily came up with the idea to rename the Arcticons Android requests file without even actually checking if it was proper xml. Derp on my part lol.

I see. Happens to the best of us :-)

> It might still be an interesting idea to have the requests file be in a format that can be more easily operated on by programs.

I agree. Now that I think about it, my personal scripts that I use for Arcticons would have been a lot easier to write if the requests file used a machine-readable format, though maybe there's a format other than XML that has a better tradeoff between human and machine readability. What about the YAML format that you use for mapping on Linux? Maybe something like this:

```yaml
requests:
  - name: OnStream
    appfilter: <item component="ComponentInfo{com.maertsno.m/com.maertsno.m.ui.main.MainActivity}" drawable="onstream"/>
    links:
      - https://play.google.com/store/apps/details?id=com.maertsno.m
      - https://f-droid.org/en/packages/com.maertsno.m/
    request_count: 14
    last_request_time: 1700051822
```

(Also, thanks for reminding me that the Matrix room exists, I've been meaning to join but totally forgot about it, haha)

dgcampbe:
Yamp would be cool, too. Perhaps someone should write a script to translate the current franken-format to yaml? I can try, I am pretty decent at python.

petlyh:
> As for the source url, check f-droid if the entry actually exists in new additions, drop if not (only keeping the playstore one) and slowly work through the backlog.

I've actually written code for one of my personal scripts that can be used for that. It attempts to fetch info about an app from the main F-Droid repo, the IzzyOnDroid repo or from Google Play (in that order, it doesn't check the other stores if found). It uses `repo/index-v2.json` for the F-Droid repos. While I personally use it to get the original icons for the requests, it could also be used to remove dead links from `requests.txt` like you suggest. I don't think reworking the code for that purpose would be that hard.

petlyh:
> Assuming Google is somewhat generous in terms of rate limiting

I actually haven't been rate-limited at all despite my script running through all the requested apps in one go without delays. So that doesn't seem to be a problem, at least for now.

> Under the assumption of keeping the F-Droid entry whenever both exist, given that those do seem to survive longer than Play Store listings...

I personally think it would be better if only the F-Droid link is kept in those cases, because they are usually longer-lasting like you said, not having multiple links would reduce the file size more, and it would be faster by avoiding calls to Google Play for apps that exist in F-Droid (which can be checked almost instantly using the index).

> As for the risk of actually keeping a mismatch (independent Apps using the same App/Packagename) absolutely no idea how likely that would be.

I would consider an app ID conflict happening by chance to be extremely unlikely unless it was a *really* generic one like `com.example.calculator`.

dgcampbe:
Should this be moved to an issue?

petlyh:
> Should this be moved to an issue?

Probably. Sorry for completely derailing the comments section of your PR. I've opened an issue here: #1831

dgcampbe:
No problem. I will close the PR since it needs more working.

PR #1824, opened 2023-11-22
Android studio svg to xml convert

Kaiserdragon2:
Convert Svgs at build time to xml drawables with android studios built in converter
add xml icon support for light and dark(disabled for now)
fix clock color in light
move some drawables (drawables that aren't icons are now all in one place /drawables for each version)


Donnnno:
Dark/Light are not XML because some popular launchers don't support that format. I'd rather keep it that way for now.

Kaiserdragon2:
I can disable that option.
Could you give me some examples which launchers don't support it. i would like to test these.

Kaiserdragon2:
Mostly generating the xml icons is much faster then the png when generating the release also the file size is much smaller. Generating all 8000 icons for the you version takes on my pc arround 20 seconds for the png it takes for just a few icons over a minute.

Kaiserdragon2:
i don,t even think that xml icons will make any problems on any launcher because the drawables are handled by android it self in most cases.

Donnnno:
Probably gonna postpone this till MS Launcher gets their shit together

xprateek:
better to create new dev branch soo new variant SVG → XML  keep updating during Christmas time.

AS Microsoft Being Microsoft they only change when required....

PR #1825, opened 2023-11-23
AirQualityWA, some cleanup of appfilter.xml

MikeTheGreat:


Kaiserdragon2:
> Any thoughts on making the usage of multi word names more consistent ? Be it the order or the actual splitting. Or doesn't matter, you'll find it anyway by searching ?

It doesn't really matter, and most of the times it depends on the naming of the actual app. Based on this the appfilter is generated and only copied over. i think it would make this process more complicated with no real benefits.

PR #1828, opened 2023-11-23
BIG CHANGE: Sorted AppFilter.xml

MikeTheGreat:
HIS IS A BIG CHANGE!!!

I wrote a Python script to sort all the entries in AppFilter.xml, because having stuff out of order was bugging me.

That said, it's basically impossible to hand-verify that everything is still in there, so feel free to reject this if you want to.

I'll upload the Python file in the next commit in case you want to see and/or use it.


Donnnno:
Hey that's cool!

But also a bit unnecessary. :S Our preparerelease.py script sorts the appfilter every time on release already.

MikeTheGreat:
So that means I can put my entry at the very top, instead of finding the right spot for it?  Nice! :)

I'll go ahead and close this, then.

Donnnno:
Yes exactly! No need to do any effort finding the alphabetically correct place for them.

PR #1835, opened 2023-11-25
Add support for Meizu phones

supertsy5:
- Add support for Meizu phones


Donnnno:
Many changes happening here that shouldn't be happening :s

PR #1841, opened 2023-11-26
More icon(s)

MikeTheGreat:


Donnnno:
Thanks for the icons :)
Please remember that less is more!

Also keep the contributing guide in mind when you're creating icons, I found lots of inconsistencies while checking this one.
Things like:
- weird corners
- double lines
- documents that were too big.
- too small elements


Scherm­afbeelding 2023-12-22 om 12 58 24

PR #1861, opened 2023-12-13
Fix presumably broken component string

bsvka:
Haven't tested, only noticed while going through appfilter.xml for something unrelated.


Donnnno:
thanks!

PR #1864, opened 2023-12-20
Fossifyication

TotallyAvailable:
Left simple_calendar and calendar_simple_ and some generics (calculator, clock, contacts, phone, messages, recorder) as is.

~Newly required drawables:~ All covered by existing for now
- ~fossify_camera~
- ~fossify_draw~
- ~fossify_filemanager~
- ~fossify_flashlight~
- ~fossify_gallery~
- ~fossify_keyboard~
- ~fossify_launcher~
- ~fossify_app_launcher~
- ~fossify_music_player~
- ~fossify_notes~
- ~fossify_thank_you~ -> renamed to a more freely usable 'thank_you'

Might just be a thing of reusing the existing 'simple_' for some or all of them under the new name to allow for easy replacement later.

Things might still change but that's easy enough to fix.


Donnnno:
Do their icons actually look different, or are they just the same?

Donnnno:
Since the icons are identical right now, I see no reason of adding the same icons again. It's best to use the same icons from the simple apps instead of adding duplicates.

Issue #1753, opened 2023-10-20
Multiple Icons for Single App. (Based on System Theme)


BaruaAtish:
Hello, I want to help contribute to Arcticons by submitting icons I create for an app called "DualWallpaper". The issue is, DualWallpaper changes icons based on system theme. I have created icons for both versions but do not know how to name and add them to the others folder and how to list them in the xml. Any help would be appreciated. Thank you!

Donnnno:
Have you tried getting it with Icon Request? https://github.com/Kaiserdragon2/IconRequest/releases

Maybe it's different when you change the mode; otherwise, automatic switching isn't possible.

We can still add both tho!

Kaiserdragon2:
I have tested it out and it is possible to do as you can see with the following debug apk.
[Debug-apk.zip](https://github.com/Donnnno/Arcticons/files/13259294/Debug-apk.zip)
But the current setup doesn't allow for this kind of build and there need to be some file manually moved and renamed.
I don't know if it is worth to change the script to allow for such builds.

Donnnno:
As far as I know, this is only possible when the dev uses different activities for the icons. If not, it's not possible for the icon pack to switch between icons.

Issue #1679, opened 2023-08-31
[You] Custom picked app icons in launchers aren't dynamic


Donnnno:
For example:

I've changed the middle icon here, and chose another dynamic theme style.

![image](https://github.com/Donnnno/Arcticons/assets/31142286/ac7d16bd-3c8c-426d-b7f2-11e0dfd25a93)

Probably external, but might be worth it to contact other devs.
Link to Nova Discord support thread. 
https://discord.com/channels/466300922767736844/1146750269535821866.

Donnnno:
Expected behavior in Nova, but they might change it soon.

ThatOneCalculator:
Can't repro in kvaestito, external

Donnnno:
Thanks for checking, every launcher probably has its own implementation. Kvaestito has lost of cool features for icon packs, so that explains it probably.

PR #1872, opened 2023-12-24
Fixes and Extended Coverage

TotallyAvailable:
- Icon binding is apparently case-sensitive.
(While uncommon and potentially problematic, still part of Google's official documentation. Not sure how the script likes it ~given that my original commit must've been (but apparently wasn't...) uppercase~)

- Fix for DataBackup, Radar & primitive ftpd

- Last commit relates to #1753 / also reduces potential issues related to #1679

- As I currently can't think of any other app riding the custom Icon wave with more than one fork, that might be it for now.


Donnnno:
Damn, you must have been busy with this one!

PR #1890, opened 2023-12-29
OpenMoji - Add 25 Black OpenMoji

Saijin-Naib:
Trial run for adding OpenMoji, named in the format openmoji_%emojipedia-name%.svg

These are currently being placed under the Black directory, though I'm not sure if they should not rather be in Other or perhaps somewhere else.

Guidance absolutely appreciated.


Donnnno:
I've looked at it, but I'm not sure if the OpenMoji set fits with the Arcticons design language.

Some factors that affect this:
- different document size, OpenMoji's 72px, Arcticons 48px. (Arcticons has a somewhat denser design)
- The design of OpenMoji has lots of open shapes.
- Also noticed some inconsistencies in the use of fill elements and different line widths

Saijin-Naib:
Hmm... are these things the script can help address to clean them up? 

I did a cleanup manually in Builder, but I may have missed things.

Kaiserdragon2:
it might be possible to write a script that scales them down and fixes line width issue's and possible some fill things. they could also be automatically be named

Saijin-Naib:
What fill problem is happening? All of them, or a specific one? 

I'll see about manually addressing the above concerns and sending a screenshot. 

I think even right now they look nicer than colored system emoji in my laumcher, so i'm hopeful they can be tweaked more to fit better. 

I will not be helpful for scripting, unfortunately...

Kaiserdragon2:
i think i will look into scripting something seems like a fun thing to do.
i don,t know about the fill @Donnnno  mentioned some fill elements. While it is possible to script it to use the outline it might not look right.

Donnnno:
> Hmm... are these things the script can help address to clean them up?

Maybe for some quick conversion, but I do prefer human made edits for the best results (if that makes sense).

Saijin-Naib:
Do a first pass with the script and tweak from that good baseline? Makes sense!

Donnnno:
Exactly!

Kaiserdragon2:
[Openmoji_conv.zip](https://github.com/Arcticons-Team/Arcticons/files/13799012/Openmoji_conv.zip)
here is a little script that converts the icons and names them.
You just have to run the script in the same folder as the input folder where the icons are located

Donnnno:
I'm still not sure about the style of the icons, it just feels a bit off

Saijin-Naib:
Do you happen to have a feel of what might match better? Twemoji? 

I think they look better than not having them, but I am also no graphic designer 🤷

Donnnno:
merged with some additional tweaks in #1990

PR #1891, opened 2023-12-30
Common Luxembourg apps

LukeVader-IV:
A set of commonly used applications in the Grand Dutchy of Luxembourg


Kaiserdragon2:
The Icons look great, but there are some things that need to be changed:

- your document unit is set to mm instead of px 
- we only need the White version 
- they should be placed in the "other" Folder

LukeVader-IV:
Thanks, i hadn't realised the correct location for the files, nor that i only needed the white ones.
I had noticed i had used the wrong scale shortly after opening the MR

For the S-net banking app, i couldn't only use outlines, i had to use filled shapes for the bridge pillars. The sharp edges on the bottom are a necessary part of the bank's branding. I hope that's fine.

LukeVader-IV:
i added an icon that is supposed to be used for a multitude of apps: Payconiq
It's a digital payment system that is common in Luxembourg, and i've seen in in belgium and NL too.

I think they can all use this same icon, as logically, only one should be installed at any time.

The following are examples of specific apps for payconiq:
- Payconiq - BGL BNP Paribas
- POST Payconiq
- BIL Payconiq
- Payconiq by Bancontact
- Raiffeisen Payconiq
Many other banks in BENELUX use this, but integrate it into the standard banking app (like Speurkees or ING)

Kaiserdragon2:
> Thanks, i hadn't realised the correct location for the files, nor that i only needed the white ones. I had noticed i had used the wrong scale shortly after opening the MR
> 
> For the S-net banking app, i couldn't only use outlines, i had to use filled shapes for the bridge pillars. The sharp edges on the bottom are a necessary part of the bank's branding. I hope that's fine.

the ions should only be strokes i think it quite doable for the pillars and they need the round edges


- the unit is still mm not px 
- to apply the icons automatically you need to add some code to the appfilter you can gather it with https://play.google.com/store/apps/details?id=de.kaiserdragon.iconrequest
- the filename should be lowercase without special characters

Donnnno:
Hi, thanks for your contribution. I've made some adjustments to make it fit more in Arcticons:

Scherm­afbeelding 2024-01-07 om 15 46 04
Scherm­afbeelding 2024-01-07 om 15 45 58
Scherm­afbeelding 2024-01-07 om 15 45 52
Scherm­afbeelding 2024-01-07 om 15 45 46

Try to keep the letters with double lines to a minimum, it makes the icons way more cluttered than they need to be.

Great work!

PR #1899, opened 2024-01-01
Add Icons for various Danish apps

Limfjorden:
Custom Icon for
- DR nyheder
- Pant
- Planday

"Vejret DMI og YR" uses simpleweather.svg as the icons are nearly identical.


Donnnno:
Thanks for the icons, they look great.

Just fixed some small things, like sizing and removing some random anchor points

PR #1901, opened 2024-01-04
added icon for 5K parkrunner results

AxFo:
My first attempt at creating an icon, please feel free to give feedback ;-) 
Thanks for the great icon pack!


Donnnno:
Hi, great work! Just fixed some minor things:
- I noticed that you used lots of capitalization with the names, capital letters make android apps crash, so fixed that.
- If a name starts with a number, there must be a _ underscore before it.
- Some of your sizing was kinda small, check the default template files to get an idea of it.

PR #1905, opened 2024-01-07
New icon added - iob_mobile

Sande-Santhosh:
Icon Name - IOB mobile


SandeSanthosh:
added activity name, added name in contribution list and changed icon name to iob_mobile

PR #1908, opened 2024-01-07
Added chesscraft icon

dgcampbe:
It's not perfect, though
chesscraft
chesscraft


Donnnno:
My dyslectic mind read cheesecraft for some reason 🙃

dgcampbe:
I would download an app called cheesecraft.

Donnnno:
It's a difficult one, but I guess it works :-)

PR #1914, opened 2024-01-09
Entries from requests.txt for which icons exist already

alchemiker:
### INFO: I somehow failed at creating a new accurate branch without my previous commits, please discard the commits before 09.01.2024, they are in another pull request of mine already ###

No new icons, only added those to appfilter from requests that already have an icon in the arcticons base already. Those are often language versions, renamed apps, apps from other places than the Play Store and thus with slightly different app pathes, flavours like "pro" or "lite" with no difference in icons (at least in black-and-white-rendition) - stuff like this.

I made a search between appfilter.xml and requests.txt file to find apps that have the same name of the launch activity. I looked up how the apps are called and if the names were different or something else was suspicious I looked the apps up on the internet, and compared their icons with the ones in the Arcticons app.

So far I came to the letter F (update: L). To be continued... but if you are going to release a new version somewhere soon, you can merge, too.

The aim is that more people see Arcticons Icons on their screen straight upon installation of the icon pack without the need to search for them, and not to waste the time of contributors by designing icons that already exist.


alchemiker:
@TotallyAvailable Yeah, I think we must find a way to minimize the workload for the long-term, but I think we should rather open up a thread in the discussions pane for this issue. I'm just going to work through my personal list in this pull-request.

We have to find a solution that works for both Donno and the icon creators.

Donnnno:
Wow, thanks for going through the requests!
Is it possible to rebase and add all the appfilter changes, I can't merge it at this stage, because of conflicts sadly.

alchemiker:
@Donnnno how do I do this? I edited the file in GitHub here in the browser...

alchemiker:
I did delete all the svgs, but there is still a problem apparently. Do I have to go through the appfilter file and delete my 28 entries from the other commit? Or make a fresh new updated branch and add my 100+ commits from from this branch to it? 🤔 I was hoping that Git could help it. 
@TotallyAvailable that's something only people with write access to the directory here can do.

petlyh:
@alchemiker I cloned your branch and did a local merge using `git rebase`. I can't push the merged branch since I don't have write access, but these are the appfilter entries that are causing the conflict:
- Domino's
- Google WebView DevTools
- Recorder

Undoing just your changes to those entries should allow an automatic merge if you want to avoid merging on the command-line. You can restore the icon files since those aren't causing the conflict.

Donnnno:
oof, lots of merge errors.

I've added all your changes in a newer commit.

I'd suggest using different branches for different PRs, that avoids merging errors like this.

Also GitHub desktop is really easy to use too :)

alchemiker:
Thank you very much for your patience, this whole story was definitely a lesson for me to make new branches in the future 😅 I am very sorry for all this mess! I'll try to make it better next time. 
Do I still have to do something with this branch or is everything finished here now?

Donnnno:
No worries, I also made a mess of it with my attempt to rebase 😆.
I think you can delete this one now.

PR #1917, opened 2024-01-11
Fixes and Additions

TotallyAvailable:
### Fixed
- Content Shell (added entry)
- Neo Backup
- Oscilloscope
- OSS Weather

### New
Name | SVG | Entry
 --- | --- | ---
Android Makers | x | x
AppDash | x | x
Black and White | x | x
Camera File Copy | x | x
Circledient | x | x
Daum | x | x
Free Browser | x | x
Hanping Popup | x | x
Ikon | x | x
Immersive Chinese | x | x
Kernel Flasher | x | x
Korean Learners Dictionary | x | x
Lapse 2 | x | x
LuX (Green/Blue/Yellow) | x | x
M3U | x | x
Meshtastic | x | x
Nate | x | x
PRISM Live | x | x
Sideband | x | x
Tibber | x | x
Xenia | x | x
네이버 메모 | x | x
대한민국 구석구석 | x | x
시리즈 | x | x
치지직 | x | x
티스토리 | x | x

- already covered lux_red by lux_icons
- moved Tibber from reusing 'Electron' to my new (slightly reworked since #1900) masterpiece.
- Assuming I didn't mess up something big while digging through the mess that keeps growing and none of the Icons require more work from you than you just doing them from scratch... expect bigger PRs quite frequently continuing with my own list initially.
- The Lux_* ones were done with the wrong width a while back - "dirty fix afterwards"...might re-do them while adding the Gold/Black versions.
Should the sizing actually be wrong (I'm sure I checked more than 1 full size Icon as comparison) let me know for the future. No holding back with the feedback, my list is rather long.
 (the Content Shell Icon could've just been done right by chance after all)
- Why now ? I might've run out of "can be covered by existing" and while mentioning my non existing creative abilities in one of my first PRs here...might as well give it a more serious try.
- Still playing around with how to make the PR look
- Totally going to sort the entrys properly next time 👍
- Also 2 more things
  - xml:space="preserve" <- keep or remove
  - fill-rule="nonzero" <- keep or remove like "evenodd"

Also thought about something like this to be added to the contribution guide.
Maybe even in some form directly in the readme. Just thinking
```
1. Take a look at the current raw [Appfilter](https://raw.githubusercontent.com/Arcticons-Team/Arcticons/main/other/appfilter.xml).
   - Search for the App or Packagename
     - If neither returns a match continue with step 2
     - If either or both appear in the filter, the Icon likely already exists and an app update changed the activity name required for the Icon to be bound.
       - **You can now either:**
       - [Open an Issue](https://github.com/Arcticons-Team/Arcticons/issues/new/choose) and tell us about it (no fork or file editing required), information from step 5 allows for a quick(er) fix.
       - skip steps 3 and 4 to update the appfilter.xml with the information you gathered in step 5 **below** the existing (broken/outdated) entry.
```


Donnnno:
Fantastic work on the icons! Just did some small tweaks on the sizing :)

PR #1923, opened 2024-01-14
Requests that already have an icon, but have different package name

alchemiker:
Made a new script which looks for doubles in activity names between appfilter.xml and requests.txt (props to ChatGPT for being patient with me), which runs only 17 seconds and also outputs the names of the entries at glance, so one doesn't have to look up the names anymore.

Added doubles (same app name, different package name, same activity name) and country variants. Didn't add variants like "pro", "beta", "HD" etc., for they have sometimes different icons.

### DIFF
--- a/other/appfilter.xml
+++ b/other/appfilter.xml
@@ -1114,6 +1114,7 @@
 
 	<!-- Alfa Bank -->
 	<item component="ComponentInfo{ru.alfabank.mobile.android/ru.alfabank.mobile.android.splash.presentation.activity.SplashActivity}" drawable="alfa_bank"/>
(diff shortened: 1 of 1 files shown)


Donnnno:
Wow that's super helpful, mind sharing the script?

alchemiker:
@TotallyAvailable in defense of my script I can say that all errors are human-made by me, I only used the script to find doubles, the addition to the file was done by hand by me 😅 
Sure I can share the code, but it's still work in progress. What is the best way to share it?
And also: is there a good place where we can lead the discussion with Donno and the other volunteers about the requests.txt file and what to do about it? At the moment our discussion is dispersed between closed issues and merged pull requests...

Donnnno:
> Also before I get the next set of Icons ready Any rough estimate/plan for the next release ?

Hopefully, this or next week, but dunno yet. I got quite a busy week ahead of me.

PR #1926, opened 2024-01-15
Faster email parser

Kaiserdragon2:
This makes the email Parser a lot faster.
I also modified the Backlog filter to filter out apps that haven't had any requests the past 6 months  and less than five requests in total. This will reduce the file a bit and also clean some dead apps out.


Donnnno:
fantastic!

PR #1931, opened 2024-01-17
Alchemiker icons

alchemiker:
Added following icons:

- Bundesliga
- Cats are Liquid (2x in requests due to different versions)
- Gov.gr 
- S LIME
- Sci Calc


Donnnno:
great work! 
some small fixes
- renamed entries S_LIME & gov.gr -> s_lime & gov_gr
- resizing

PR #1936, opened 2024-01-21
Added Streak Society Duolingo Icon

SagXD:
I just achieved a 30-day streak on Duolingo, which gives us a special icon. However, it uses the same old articon icon, so I created one for the streak society icon. I'm still not sure if the entry in the 'appfilter.xml' is correct because it uses 'drawable=duolingo' on both activities


SagXD:
> Let's say you stick around and start working on the massive backlog in the requests.txt and end up picking an App with drawable="inventory" . Valid request, valid drawable name, except you are now going to replace an already existing icon for an already covered app (intentional behavior, the silent replacement of existing Icons). If you now keep the "piggybacking" of unrelated apps using existing Icons all over the place and the not that uncommon breakages due to (Main)Activity changes in mind you'll get an idea of what could happen if someone who just started using Arcticons, decides to contribute an Icon (not knowing about the existing but broken listing) without ("extensive") research. And the more you look the more confusing things can get depending on what you want to do.

Yep, it's true. I was also planning to use 'duolingo_alt,' but later I searched for it and found that there's already an icon named 'duolingo_alt.' There should be some sort of bot that checks for existing icons in pull requests.

PR #1940, opened 2024-01-22
Add icons for Lingodeer, NightOwl and Tasks.org

Elaborendum:
Tasks.org's activity name changes depending on the currently selected icon colour in the in-app settings (e.g. changing it to red turns it into com.todoroo.astrid.activity.TaskListActivity). I've added the default name, but I don't know how this would interact with this.


Donnnno:
Hi Tasks org should already use the `tasks` icon, so we wanna assign that existing one, instead of creating a new one

Donnnno:
Nice work on the icons! I've noticed a couple of double lines in the owl icon, so fixed that.

Issue #1942, opened 2024-01-23
XML error in appfilter 


Donnnno:
I wanted to test the latest build, but the appfilter corrupts when it gets sorted by the preparerelease script

Kaiserdragon2:
I will look into it later today

Donnnno:
Thanks @petlyh ! how did you find it?

Donnnno:
Fixing the line resolved the problem!

petlyh:
> how did you find it?

Old reliable `git bisect`. It's incredibly useful for things like this.

Donnnno:
> Should be another thing possibly looked for by the script like duplicates and missing drawables ? Check if every '!' has a '<' in front ?

Good suggestion, looks like @Kaiserdragon2 did some scripting magic again to implement this :)

PR #1943, opened 2024-01-24
Add check for formatting of the appfilter.xml

Kaiserdragon2:
this should highlight errors like #1942


Donnnno:
Fantastic work :)

PR #1944, opened 2024-01-24
Lets go Faster (Improve Preparerelease script)

Kaiserdragon2:
Use Cairosvg instead of inkscape to create png
improve runtime for current build from 390s to 12s on my machine

### DIFF
--- a/other/scripts/preparerelease.py
+++ b/other/scripts/preparerelease.py
@@ -1,19 +1,17 @@
 from pathlib import Path
-import shutil
 import re
 import glob
 import subprocess
 from shutil import copy2
 from typing import List
-import sys
 import argparse
 import pathlib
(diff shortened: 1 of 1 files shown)


Donnnno:
Wow, this is absolutely amazing! This is gonna save so much time!!!

Donnnno:
might wanna check this tho:

image

idk what I have to fix now 😅

Donnnno:
Just tested it, something's happening here probably:

image

Donnnno:
Also the iconpack.xml files are unnecessary, so we probably want to remove those as well :)

Kaiserdragon2:
> Also the iconpack.xml files are unnecessary, so we probably want to remove those as well :) 

Will look into some optimisation of the script

PR #1945, opened 2024-01-24
added icons

AxFo:
added icons:

* parkrunner Virtual Volunteer
* added alt-version for the clock icon, as the old icon shows an impossible time (which irritated me quite a lot :sweat_smile: The alt-version is simply a rotated version of the original one.


Donnnno:
nice!

PR #1950, opened 2024-01-26
appfilter.xml: Add "Gadgetbridge Nightly" and "Gadgetbridge Nightly No Pebble"

yetanother-user:
Note that the original icon is the same for both "Gadgetbridge" and "Gadgetbridge Nightly," the only thing that differs is the coloring.

---

Gadgetbridge:

Gadgetbridge Nightly:


Donnnno:
cool

PR #1951, opened 2024-01-26
A few fixes

dioxias:
app finder was added as appfinder, so i removed the previous version of the icon and corrected the lastest version
skiff drive had a mysterious line, removed it.
opencomicvine was out of order (was OVC, changed it to OCV)
perfect piano had a duplicate icon that was added at the same time, that is exactly the same but named Pefrect Piano. so i removed the incorrectly named one.


Donnnno:
Thanks for the fixes!

PR #1954, opened 2024-01-28
Added icons

drenat:
- [Mihon](https://mihon.app/)
- [NU Client](https://play.google.com/store/apps/details?id=knf.nuclient)

It's my first time doing this :)


petlyh:
I think `mihon2.svg` is supposed to be named `mihon.svg`.

Issue #1938, opened 2024-01-21
Better categorisation 


Donnnno:
Basically have some more categories so that the user can search the app more quickly, like:
- Google
- Microsoft
- Games
- System

The challenge would be how to do it with the automatic sorting.

Donnnno:
I've added a Google and Microsoft line in the preparerelease script, so that should be covered now.

I don't think it's a good idea to add a `game_` or `system_` prefix for everything. We should probably add it by hand if we do something like that. But we have to find a way so that the automatic sorting / adding of icons doesn't mess with those categories.

Donnnno:
I just had a thought. It still requires many manual edits, but worth a shot, maybe.

What if the preparereleasescript starts with all the normal stuff: new, calendar, emoji, brands etc. and after that searches through a different file, for example a game category.

In the game category file is a list that can be edited, so everyone can add additions to it. 

```
_2048
_20_minutes_till_dawn
_2_player_games
_2248
afk_arena
alto
altos_odyssey
angry_birds
...
```


After the script has done that, it continues with the last part, ordering everything.
That way, the specific categories won't get overridden with each release.

Kaiserdragon2:
That would be possible the good thing with this would be that the files don't need to be renamed

Donnnno:
Exactly, that was that I was thinking too :)

Donnnno:
I've added a simple XML file called games.xml to test with. However, I tried to understand how the preparereleasescript works, but just can't wrap my head around it. 🙃
So I decided to add the label help wanted, for anyone who likes to figure out how the better categorization system should work. 
https://github.com/Arcticons-Team/Arcticons/issues/1938#issuecomment-2212419913

https://github.com/Arcticons-Team/Arcticons/blob/main/scripts/games.xml

Kaiserdragon2:
I will write up something just need a bit of time.
The games.xml already helps to test it.

Kaiserdragon2:
![image](https://github.com/user-attachments/assets/a55cff00-71ed-4606-b405-a106e5c054df)

gewaleelek:
Hi, I'm not sure if I should add this here, but the folder tab contains icons for the apps called "folder in folder" and "folder widget" which probably shouldn't be there.

Would renaming them so their names don't start with "folder_" work?

Kaiserdragon2:
Yes renaming them would work.

skedastically:
[...]

PR #1965, opened 2024-02-02
Move the creation of the drawable.xml to Android studio

Kaiserdragon2:
this moves the creation of the drawable.xml and the copy of the appfilter file to Android studio.
Also it adds an option to the preparerelease script to allow keeping the the current new drawables and add new icons to it.
This allows the script to be run in between releases without the need to revert these changes to have all new icons in new.
```
usage: preparerelease.py [-h] [--checkonly] [--new] ARCTICONS_DIR

positional arguments:
  ARCTICONS_DIR  Path to the Arcticons directory

options:
  -h, --help     show this help message and exit
  --checkonly    Run checks only
  --new          Run a new Release

```

Instead of of having a category for every letter i merged them to A-Z. To have an Specific Letter you will still be able to scroll to it with the scrollbar #1938


Donnnno:
Wow, that's so useful, thank you so much!

Could you add a description of the new prepare release script, with its options here?
https://github.com/Arcticons-Team/Arcticons/tree/main/other/scripts

PR #1969, opened 2024-02-03
Delete and add to git ignore

Kaiserdragon2:
These files are generated each build and copied to the correct location
added files to clean build
removed depreciated android.defaults.buildfeatures.buildconfig=true


Donnnno:
The arcticons site depends on the drawable.xml file! So better not to delete that one.

Also there's nothing wrong with those files being in the directory, right? (because I'd rather keep them in there if you don't mind)

Kaiserdragon2:
Wasn't  aware of it yeah in this case it should stay

Donnnno:
Oh yeah, almost merged it, but then that came to mind 🙂

PR #1970, opened 2024-02-04
Add icons for Locals and GCEFCU

ZachIndigo:
Adding in two icons for the aforementioned:
- Locals (com.locals.localsapp)
- GCEFCU (com.gulfcoastfcu.gulfcoastfcu)

I have gone over the CONTRIBUTING.md file, and I believe everything is in order.


Donnnno:
Hi, congrats on your first contribution, I just checked it and the GCEFCU icon is good!

I removed the Locals icon because we already have standard icons for letters like that, in this case letter_uppercase_circle_l. I've changed the appfilter so that Locals will display it :)

PR #1975, opened 2024-02-04
Fix Categories to show all existing and add them to all icons

Kaiserdragon2:
Not Urgent


Kaiserdragon2:
> Does the current implementation actually allow for an "Alternatives" category `('_alt' + '_alt_*')` as mentioned in #1938 or would that require more/too much work.

At the moment we check the beginning of the filename i don't know if it is easily possible. I will have to look into that

PR #1980, opened 2024-02-07
Added 4 apps (Same, Vibrator, ShareX and Mercurygram) and Fixed or Reused entry of 3 apps(Mimo, Brotato and Cherrygram)

SagXD:


Donnnno:
Merging your icons into https://github.com/Arcticons-Team/Arcticons/pull/2000

I've changed some directories, while I didn't merge all requests. So that gave some errors (my bad!)

For future icons, they go into /newicons instead of /other :)

PR #1981, opened 2024-02-08
Added 1 icon

ojppe:
https://f-droid.org/packages/org.woheller69.whobird/


Donnnno:
Merging your icons into https://github.com/Arcticons-Team/Arcticons/pull/2000

I've changed some directories, while I didn't merge all requests. So that gave some errors (my bad!)

For future icons, they go into /newicons instead of /other :)


(oh and make sure that your artboard is the right size, because it was huge :)

Issue #1986, opened 2024-02-12
Restructure the 'other' folder


Donnnno:
other is kinda a undescriptive name for the folder/

- I think it should be something like newicons
idk if it's logical to have requests.txt and appfilter.xml in there then
- scripts should be a folder at the top level

Kaiserdragon2:
I think the appfilter file should be in the same path like the newicons so contribute especially new ones have one folder to go to.
The other files could be easily moved especially with the planned change for the request site, the request file must not be that accessible. The scripts at the top is a good idea.

Donnnno:
Works for me

Issue #1985, opened 2024-02-12
Further enhancements to the prepare release script


Donnnno:
We can make our fantastic script even more powerfull with these changes:

- newdrawables is not deleted upon generation

![image](https://github.com/Arcticons-Team/Arcticons/assets/31142286/29e6f8ea-47aa-41a5-b663-d0c52f5e4464)

- Motorola support: https://github.com/zixpo/candybar/pull/174
by copying and renaming appfilter to icon_config.xml we can add support for moto devices!

- the 'other' folder with empty icons doesn't update the appfilter in res & assets (this might be something only on my end)

Kaiserdragon2:
Appfilter is updated through android studio at build.
The new drawable file is needed to ensure the new category stays the same if the app is rebuilt in case of a fix

Kaiserdragon2:
I can add the the icon_config file to the preparehelper script

Issue #1831, opened 2023-11-24
Improvements to the requests file


petlyh:
(Continuation of the discussion in the comments of #1822)

Some possible improvements to the requests file:
- Move to a machine readable format to make it easier for scripts to interact with it. My suggestion would be YAML since it has a good tradeoff between human and machine readability. Example of a possible structure:
```yaml
requests:
  - name: OnStream
    appfilter: <item component="ComponentInfo{com.maertsno.m/com.maertsno.m.ui.main.MainActivity}" drawable="onstream"/>
    links:
      - https://play.google.com/store/apps/details?id=com.maertsno.m
      - https://f-droid.org/en/packages/com.maertsno.m/
    request_count: 14
    last_request_time: 1700051822
```
- Only include app store links that actually work. This would likely involve a script that looks up the app IDs of new requests in multiple app stores to see if they have the requested app. I have already written code previously that can be used for this (currently checks F-Droid, IzzyOnDroid and Google Play) and I would be willing to write the script for this if desired.
- ~~Remove "Last requested time" to reduce file size if it's not going to be used for anything.~~
I didn't realize it was being used to filter old requests. My bad.

petlyh:
I took a look at the `email_parser.py` script and it looks like the last request time is not used for [sorting](https://github.com/Donnnno/Arcticons/blob/main/other/scripts/email_parser.py#L252) but *is* used for [filtering old requests](https://github.com/Donnnno/Arcticons/blob/main/other/scripts/email_parser.py#L81). Since that functionality also reduces the file size of the requests file (as well as the number of requests), removing the last request time is probably not a good idea after all.

> request_count -> rq_count/r_count

I think it's important to maintain a certain level of human readability, so maybe abbreviations should be avoided.

> Maybe even drop the "item component=" part (appfilter: "ComponentInfo...) as the script can just "add" it if needed for comparing stuff. As for manually moving it into the appfilter with new icons...it's there, it's everywhere, at least 8 thousand times.

I'm guessing that the requests file contains the full appfilter XML entry for each request to make it easier for contributors (especially new ones) to just copy it from the requests file without needing to understand it. That's why I kept it in the example structure.

> really pushing it: play.google.com/ -> play.google/

That won't work since `play.google` links just redirect to the front page. Follow this link to see what I mean:
https://play.google/store/apps/details?id=com.donnnno.arcticons

Kaiserdragon2:
[...]

PR #1987, opened 2024-02-13
restructuring and enhancing 

Kaiserdragon2:
support for icon_config.xml  #1985
restructure #1986
request html  #1831

i have made it so that the github action pushes its generated file to the new branch and the gh-page is still on main


Kaiserdragon2:
one moment i need to update the contributing file

Kaiserdragon2:
now

PR #1990, opened 2024-02-13
>20 Icons 

Kaiserdragon2:
Includes all emojis from  #1890


Donnnno:
Great work, tweaked some things here and there.

PR #1999, opened 2024-02-15
Added 2 new apps

GaensebIume:
remotecall_service (LG vendor app)
android cell_broadcast


Donnnno:
Noticed a fill element, so replaced that with lines, also noticed that the cell icon was way out of bounds.

PR #2005, opened 2024-02-16
Fix combining the appfilterfiles & allow action only run 1 at the time

Kaiserdragon2:


Donnnno:
This thing just won't get fixed that easily, huh?

Kaiserdragon2:
It's hard to test if you need to interact with pull requests and the repo itself

Donnnno:
I can imagine, I've given you write access so that you don't have to wait for me every single time :)

Kaiserdragon2:
Now it works

PR #2019, opened 2024-02-29
7 new icons

savetier:
4ka Slovakia

Unicredit Business Smart Banking Slovakia

Diskdigger Pro

Infomaniak kMail

YouHodler

Health Insurance Slovakia

Packeta Slovakia


Donnnno:
Nice work! 
Can you use the normal icon names next time? 
Also check the line-width before uploading the icons.

Thanks :)

savetier:
> Nice work! Can you use the normal icon names next time? Also check the line-width before uploading the icons.
> 
> Thanks :)

Hi, sorry for the inconveniences, this was my first commit ever ;). What do you mean by line-width?

Donnnno:
Yeah, no worries! :-) 

The line width is the stroke thickness. The default value should be 1px (or pt) so that every line has an equal width. It's a very common thing that it isn't 1 tho, and that some strokes have different widths.

PR #2022, opened 2024-03-05
Add 5 icons

willycast:
Added: Computrabajo,Mi Personal Flow, Personal Pay, Uptodown App Store and Tiendamia


Donnnno:
Great work on the icons!
I've made some small modifications:
image

I noticed that some text elements were stretched, so I fixed that.

PR #2031, opened 2024-03-11
Add 2 Icon

willycast:
Added Citibanamex and Squawker


Donnnno:
Squawker already has an icon, and I removed the trademark from the other icon, because those are clutter :)

willycast:
> Squawker already has an icon, and I removed the trademark from the other icon, because those are clutter :)

didn't know about squawker sorry, about the other icon you're right, looks better without the trademark

PR #2074, opened 2024-04-04
Add new Beeper app icon

andersmmg:
For the new android app, com.beeper.android


Donnnno:
Already updated the beeper icon to this one in v9!

PR #2076, opened 2024-04-05
Change String for launcher support remove unused icons

Kaiserdragon2:
Strings might need some refinement


Donnnno:
Ah nice, I'll be modifying the description a bit more tho, some of the other bits could use an update too :)

Donnnno:
Something like this
Scherm­afbeelding 2024-04-17 om 15 17 11

PR #2085, opened 2024-04-11
Update appfilter.xml

Kaiserdragon2:


Donnnno:
Thank you for all your work the past month! 
Sorry that I'm not as active as usual, but personal life, work and other stuff got a bit in the way 🙃

PR #2091, opened 2024-04-12
Fix Sorting arrow move request.txt

Kaiserdragon2:
@Donnnno I moved the request.txt and the updatable.txt into the docs/assets/ folder.
there they will be served with the the rest of the site and it is simpler to link to them inside the javascript.


Donnnno:
Sounds good! Thanks for giving me a heads up :)

PR #2094, opened 2024-04-16
Add icon: wasteof for android

Cequallium:
Adds icons for com.micahlindley.wasteofmobile


Donnnno:
Nice work! 
I've made sure that the icon doesn't touch the bounds

PR #2095, opened 2024-04-16
Add extra smartlauncher support

Donnnno:
Smart launcher has some cool icon pack support.
Especially for Material You stuff.

https://docs.smartlauncher.net/work-with-smart-launcher/icon-packs

Duplicating the AndroidManifest.xml like a stupid way to do this tho...


Kaiserdragon2:
https://developer.android.com/build/manage-manifests 

This seems to be possible it should work similar to the configuration file for the dashboard where only the attributes that are different are declared

Donnnno:
Ah yeah, but the whole MainActivity block should be put into the file right? 

```
        <activity
            android:name=".activities.MainActivity"
            android:launchMode="singleTop"
            android:theme="@style/Theme.App.Starting"
            android:configChanges="uiMode|orientation|keyboardHidden|screenSize"
            android:exported="true">
```

Kaiserdragon2:
i have reduced it to this for the yuou variant and it seems to be merged correctly 
<?xml version="1.0" encoding="utf-8"?>
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:tools="http://schemas.android.com/tools">

    <application>
        <activity
            android:name=".activities.MainActivity"
            android:exported="true">
            
            <intent-filter tools:node="merge" >
                <action android:name="org.adw.launcher.THEMES" />
                <category android:name="android.intent.category.DEFAULT" />
                <category android:name="org.icontheme.CHANGES_WITH_MATERIAL_YOU_COLORS" />
                <category android:name="org.icontheme.FALLBACK_TO_THEMED_ICON" />
                <category android:name="org.icontheme.CHANGES_WITH_DARK_MODE" />
            </intent-filter>
        </activity>
    </application>
</manifest>

Donnnno:
Oh nice! Can you merge it into this commit or create a new one? :)

Kaiserdragon2:
> Oh nice! Can you merge it into this commit or create a new one? :)

Pushed to it

Donnnno:
thank you!

DeltriDev:
Nice, I use smart launcher, and that's pretty neat

Donnnno:
So in my initial testings, the extra lines caused the features to not work.
I have to take a look again in the future what causes it.

Kaiserdragon2:
Just did Some Testing with our current release and these smartlauncher feature's were already working.
After some digging i found out that this line (already in the manifest) >
```
<!-- Lawnchair Launcher -->
<intent-filter>
      <action android:name="app.lawnchair.icons.THEMED_ICON" />
      <category android:name="android.intent.category.DEFAULT" />
</intent-filter>
```
makes the feature also work.
It also categorizes the iconpacks in the Lawnchair-Launcher as "Themed Icon Source" which the normal and black Arcticons are not.

Kaiserdragon2:
> so this commit was unnecessary?

I think the smartlauncher devs just changed something and now it works without these changes. or at some point we added the lawnchair thing that made it work.
but now it seams not to be necessary.

Kaiserdragon2:
That's good to know then we can create an appropriate config for each version. One thing what i am not sure about is the day and night version that isn't material you, but needs to be refreshed on Day/Night Mode switching.

Donnnno:
I think so but, will have to double-check that. 
This seems good for now.
It took a year, but we can finally merge this 😅

PR #2116, opened 2024-05-08
Added ParentSquare icon

AblertARock:


Kaiserdragon2:
Could you also provide the appfilter code you can gather it with https://github.com/Kaiserdragon2/IconRequest

PR #2119, opened 2024-05-10
Memories icon

bonswouar:
Please don't merge this 😄 
I'm not a designer at all and I've never used InkScape before, I couldn't get this circle thing right... But if anybody a bit more skilled is motivated to update it, here's my dirty work (:

This is for [`gallery.memories`](https://f-droid.org/packages/gallery.memories/) btw, adapted from [their original svg](https://github.com/pulsejet/memories/blob/master/img/app.svg)


bonswouar:
Hmm I don't know what I did wrong but compared to other PRs it seems my image is very small, I've followed the instructions for InkScape though (and I can see it's effectively 48x48 with 1px scale), what did I miss?!

Kaiserdragon2:
> Hmm I don't know what I did wrong but compared to some other PRs it seems my image is very small, I've followed the instructions for InkScape though (and I can see it's effectively 48x48 with 1px scale), did I miss something?!

Your file is 48x 48 that is correct the only difference is that in your file the heigth and width for render is also defined with 48 if you would save it as plain svg this isn't in there and the image gets rendered bigger on github if compared to the inkscape svg.

But this isn't necessary.

bonswouar:
> > Please don't merge this 😄 I'm not a designer at all and I've never used InkScape before, I couldn't get this circle thing right... But if anybody a bit more skilled is motivated to update it, here's my dirty work (:
> > This is for [`gallery.memories`](https://f-droid.org/packages/gallery.memories/) btw, adapted from [their original svg](https://github.com/pulsejet/memories/blob/master/img/app.svg)
> 
> this is not all bad, but some things need change: there shouldn't be any fill try drawing around the outlines of the shapes from the original image
> 
> we also need the appfilter code you can gather it with https://github.com/Kaiserdragon2/IconRequest

Got it thanks!

Donnnno:
Please take a look at our contribution guide! Your icon doesn't match the style right now, and cannot be merged. 
The SVG is also very cluttered

image
image

dgcampbe:
@Donnnno *nudge* since this is an older PR :D

Donnnno:
image
 done

PR #2120, opened 2024-05-12
new, alternate, and updated icons

akp37:
bitwarden authenticator
bleacher report
flut renamer
hot stuff
notesnook
private dns quick toggle
quranapp
retro mode neon
touch the notch


Donnnno:
Nice work! I cleaned up the shapes:

image

Also noticed the double icon for Retro Mode because it's the same, it's better to use the already existing one in the appfilter.

PR #2121, opened 2024-05-14
Four new icons

aloxe:
Addblock plus
icsx5
MH Share
Yen a mare des squares


Donnnno:
Nice work!
Two things:
- APB already has an icon
- the ICSx5 icon felt a bit sluggish, so I recreated the icon: 


image

Donnnno (approved):

PR #2125, opened 2024-05-17
four icones with vélo

aloxe:
vélib bike in Paris
veloh! bike in Luxembourg
Zebullo bike in Reims
Prague on bike


Donnnno:
Nice work!

PR #2127, opened 2024-05-20
change month limit and min request to reduce backlog

Kaiserdragon2:
this reduces the request file down to 12000 and request with less than 4 will be purged after two month tho multiple people need to request an icon to have it stay


Donnnno:
thanks 😌

PR #2135, opened 2024-05-23
added one icon

Rescue9:


Donnnno:
Hi, thanks for your contribution!
A few things to keep in mind for the next time:
- file names can't have capital letters inside them. 
- Use the template files to start with, the document was way bigger than 48x48px 
- try to keep it as minimal as possible: I've remade your icon to fit more into Arcticons

image

PR #2137, opened 2024-05-26
léto s dečkem

aloxe:
one icon for the summer app léto s dečkem


Donnnno:
I've recreated the icon because it was a bit messy:

image

Thanks for the effort tho :)

Issue #2136, opened 2024-05-26
[help] Icon appears in website but not in my homescreen


PanDabStudios:
Info:
Pixel 6 128
Calyx OS
Nova launcher 8.0.18 (aurora store).
Articons material v9.2.4 (fdroid)

All icons work fine, except for a few. Here is some info on two of them:

Claro Flex (aurora store) - Appears as an icon on the website and source code, but not on my home screen
Sound Meter (fdroid) -  Appears as an icon on the website and source code, but not on my home screen

dgcampbe:
Sounds like the appfilter entries for those apps needs to be updated. You can acquire the XML snippets needed to update then using an app like [IconRequest](https://github.com/Kaiserdragon2/IconRequest).

PanDabStudios:
doing a quick crtl + f in the appfilter.xml file, I don't seem to find any mention of the claro flex app whatsoever [there are a ton of other claro apps declared tho], so maybe it is missing.

Now the sound meter logo seems to be pointing to com.gamebasic.decibel, which I believe is another sound meter available in the google play store [non free, with ads, and the developers website is some sort of chinese mess], maybe the way to go with this one is to also add the app ID for the free one that I am using? 

 Here is the code I got from IconRequest for both of them: 
```
<appfilter>

	<!-- Claro flex -->
	<item component="ComponentInfo{br.com.claro.flex/br.com.claro.flex.features.splash.ui.SplashActivity}" drawable="claro_flex"/>

	<!-- Sound Meter -->
	<item component="ComponentInfo{com.splendapps.decibel/com.splendapps.decibel.MainActivity}" drawable="sound_meter"/>

</appfilter>
```

wwwwwwari:
Icons should now be fixed in the latest release 🙂

![Screenshot_2024-06-02-18-14-43-65_3822ac0e29cfec8970059bb21d01f05e.jpg](https://github.com/Arcticons-Team/Arcticons/assets/89649458/64611732-90f6-4904-aa85-91d59544054f)

Donnnno:
thanks for the fix @wwwwwwari !

Issue #2117, opened 2024-05-09
comdirect apps + sonos icon updates


ChaosNicro:
Hey there,

Just wanted to note that the comdirect banking-app has an icon, but is not auto-detected correctly. The Tan-app for it does not have an icon yet but the auto-generated path-tracing is very good and might be usable as-is.

I also just noticed that Sonos does not apply it's icon automatically. 

Here are the relevant info-blocks:

comdirect
[de.comdirect.app/de.comdirect.core.feature.splash.](https://de.comdirect.app/de.comdirect.core.feature.splash.SplashActivity)[SplashActivity](https://de.comdirect.app/de.comdirect.core.feature.splash.SplashActivity)
https://f-droid.org/en/packages/de.comdirect.app/

photoTAN
com.comdirect.phototan/de.comdirect.phototan.feature.splash.impl.SplashActivity
https://f-droid.org/en/packages/com.comdirect.phototan/

Sonos
com.sonos.acr2/com.sonos.passport.ui.launcheractivity.LauncherActivity
https://f-droid.org/en/packages/com.sonos.acr2/

Kaiserdragon2:
I have added the comdirect app and the Sonos one to the appfilter file these should work in the next release.
For the PhotoTan it might be best to request it through the Arcticons app or draw it yourself.

wwwwwwari:
I added photoTAN in #2128 since I think it's still available at [Google Play](https://play.google.com/store/apps/details?id=com.comdirect.phototan) with the exact component string.

ChaosNicro:
> I added photoTAN in #2128 since I think it's still available at [Google Play](https://play.google.com/store/apps/details?id=com.comdirect.phototan) with the exact component string.

Thanks for picking this up, looks good. Though I would suggest keeping the "C" as an outline instead of a single line, that way it can match the main app.

wwwwwwari:
I can change the shape of the C to look like the original icon instead of using the Arcticons font. The changes won't be in this release, though.

PR #2138, opened 2024-05-27
Small icon and appfilter changes

wwwwwwari:
This PR includes mostly minor changes

1. Add several variants of Thai TV apps to use 'thai_tv.svg' icon    
2. Adjust Comdirect PhotoTAN's icon to use outline instead of the default font - #2117
3. Add missing entries for Claro Flex and Sound Meter in the appfilter.xml (icons already exist) - #2136


Donnnno:
image
 
More minimalism™️!

wwwwwwari:
> image
> 
> More minimalism™️!

Since the OP of the linked issue wanted them to look alike, should we also update [comdirect.svg](https://github.com/Arcticons-Team/Arcticons/blob/main/icons%2Fblack%2Fcomdirect.svg) to match?

Donnnno:
Hmm that C is pretty big, so that would feel maybe too empty then?

wwwwwwari:
It'll be kind of empty, yes, but I kind of agree with the OP of that issue that the 2 icons should look the same since it's the same company's apps. I'm thinking we could do one of these:

1. Copy the C of comdirect_phototan.svg to comdirect.svg - the comdirect icon will become kind of empty as you said
2. Copy the C of comdirect.svg to comdirect_phototan.svg - the phototan icon will look not minimal enough (which is what you just fixed)
3. Do 1 or 2 and add the other design as alt icons?

I'm leaning to just go with the outlines for both of them and add the thin line designs as alts - thoughts? 🤔

PR #2162, opened 2024-06-10
Some Icons

Kaiserdragon2:


Donnnno:
Nice work! 
Just make sure to update your default square templates :)

Kaiserdragon2:
> Nice work! Just make sure to update your default square templates :)

Thanks for the reminder totally forgot about that

PR #2169, opened 2024-06-25
Added slither.io icon

andersmmg:


Donnnno:
Hi! Looks like you've expanded the shape.
image

All icons should have only line elements

andersmmg:
Ah dang it I swear I undid that. I'll have to redo it sometime I guess.

wwwwwwari:
So, I've tried to redraw this icon. It's not going to be exactly the same as what's submitted, but it should still look close enough?
slitherio

Donnnno:
thx @wwwwwwari !

PR #2172, opened 2024-07-06
Added icon for the new version of Innertune

thecloakmaster:
The newer version of Innertune just has a different package name so the older icon has been reused.


Donnnno:
Looks like someone else added it too in a commit, so closing this one.
Thanks for the effort :)

PR #2179, opened 2024-07-07
Appfilter updates

Kaiserdragon2:


Donnnno:
Thanks! You're amazing for adding these :))

PR #2185, opened 2024-07-13
100+ new icons and a few fixes

wwwwwwari:
This PR includes:
* Add 100 new icons including some popular Laos apps and requested apps
* Add 11 new apps using existing icons
* Fix 3 previously submitted icons (Meb and Line Dictionaries)

**Note:** I've reinstalled my OS and I'm not sure if I set the Inkscape's settings wrong somewhere, but the XML of all new icons have their stroke-width set to .265. However, the icons themselves look correct with that value, and forcing the values to be 1 makes the strokes thicker than they should be, at least on my PC.

Please let me know if I should change something somewhere.


Donnnno:
Omg 🤯

Kaiserdragon2:
Wow that is an huge amount of icons

wwwwwwari:
Just want to try to hit 10k before my semester starts 😂

Donnnno:
Found some errors! (and am fixing then) 

Almost 1px 😅 
image

Also, I found some errors in your rounded square. it seems to be a bit off:

image
image

Circles 
image

wwwwwwari:
I think there's really something wrong with my InkScape settings lol. I can look into it myself tomorrow (and maybe submit unoptimized SVGs straight from InkScape?)

I exported all of them under the same settings so all submitted SVGs probably have incorrect values.

Donnnno:
That's really weird, there must be some kind of bug then? It might be worth letting the Inkscape team know if there's an issue with the export settings.

Donnnno:
Nevertheless, amazing work. Really! I can't thank you enough for doing such incredible work! :)

wwwwwwari:
Thanks for going through all of them too. That's gotta be tiring 🫠

Donnnno:
I got a script that imports all icons into Illustrator with their artboard names, so that makes it easier, haha.

PR #2190, opened 2024-07-15
Contributer image

Kaiserdragon2:
I have found out what the internet permission was for.
It was used to show the images from Contributors in the About Tab.
I created a solution that downloads the images at build and puts them in the assets Folder.

The original contributor.xml is moved to the generated Folder(dam this needs a better name) this should be the one that gets edited the one in the app/.../xml folder will be generated at build time.

This allows showing the images without having the internet permission.


Donnnno:
Fantastic, I figured that out a while ago, but didn't feel the need to add the internet permission just for such a small thing.

Great that you've found a way to add it back! 🎉

Donnnno:
is this ready? :)

Kaiserdragon2:
No need for a little tweak to start with the build but i can do this later today

Donnnno:
Oh, no rush, at all. The new release is at least a few days away (I hope)

Kaiserdragon2:
Now it is ready

Donnnno:
You're the best!

PR #2191, opened 2024-07-17
Dynamic clock the return

Donnnno:
I was wondering where the dynamic clock went. So I did some digging.
A few things are happening that make it not work as it should:
- the preparereleasescript was removing the appfilter entry (fixed that, kinda messy tho)
- Material You builds have the problem where it generates a generic "clock" icon from the SVG every time, so the app prioritizes that before the dynamic one.
- The same applies for D&N, but haven't tested and made the clock parts for that yet.


Donnnno:
Lawnchair and neo both seem to crash with the layer-list based clock icons :s

Donnnno:
I don't think it's worth it because the lack of launcher support. so closing this

PR #2193, opened 2024-07-18
Remove dbna icon

autinerd:
The DBNA app is getting removed due to the end of the platform (https://mag.dbna.com/leben/tschuess-dbna-community-49818 in German)

So the icon is no longer needed.


Donnnno:
ah, idk if we did that before, but sure!

PR #2196, opened 2024-07-19
Add 4 icons and Update appfilter.xml

Npepperlinux:
Add icons:
- CubeX
- DPD
- Skred
- Kredivo

Update appfilter.xml:
- Linphone
- Proton Pass
  - F-Droid version
- Proton VPN


Donnnno:
Nice work (protonVPN was added in another PR too, so removed that one again)

PR #2198, opened 2024-07-20
Add 6 icons and update appfilter

Crazy-Unicooorn:
Here are 6 icons by me, let me know if there is anything wrong!

I also add 1 item to "Proton VPN" in appfilter, because my app is not included somehow 🤔 

image


Donnnno:
Thanks!
I've resized it a bit:
image

PR #2199, opened 2024-07-20
One new icon + 2 intent filters

alchemiker:
Hello, I added two intent filters to AnySoftKeyboard (ASK) because the language packs installed from the PlayStore are permanently showing and thus need an icon. I put them under the regular ASK icon, because it would be a nuisance to create an icon for every single one of them -- flags don't transfer good to black&white, I'd had to look up language codes, and anyways nobody needs to access these "apps" manually from the app list, because they are in reality just extensions for ASK.

I also added a new icon for [Text Tools](https://f-droid.org/packages/com.corphish.quicktools/) from FDroid, with entry in appfilter.xml and with the svg icon in newicons folder.

My github online was acting up so I couldn't see a preview of my changes, but I hope it turned out fine.


Donnnno:
Thanks!
I noticed two things: 
- the appfilter was set to CRLF instead of LF
- your square icon was a bit off, it used the old template (the new one is rounder). And had some extra unneeded anchor points.
- 
image

PR #2215, opened 2024-07-25
Added some apps

undefynedd:
…d apps. There are two apps named Akebi though, and I am not sure if there is anything I should do to change that.


undefynedd:
> I think both Akebi are the same app, right? The component infos and even the icons seem to be the same.

Yeah, it seems so. Weird that it doesn't show up in Arcticons though?

Donnnno:
Congrats on your first contribution, nice work on the icons!

Just fixed some small things, like line-width & sizing

PR #2219, opened 2024-07-27
Build release action

Kaiserdragon2:
This adds two workflows:

1. Build debug apks and upload as artifacts can be triggered manually
2. Build Release apks and create a Release with Release Notes Image and Assets

For the signing of the APK to work you need to create these secreats
image

The KEYSTORE holds your .keystore/arcticonskeystore.jks file as base64 encoding
On linux you can convert it like this `base64 -w 0 arcticonskeystore.jks > key.txt`
The KEY_ALIAS holds the name of the key to sign the apks
The KEYSTORE_PASSWORD holds the password to access the keystore
And the KEY_PASSWORD holds the password for the key

To test if it works correctly you can run the debug apk workflow and check the output for errors
if it runs successfully you can download the artifact and check if the apks are signed with your key 
Therefore  you can use https://f-droid.org/packages/com.smartpack.packagemanager/

### How to create a new release

1. Have everything ready as you normally would have (version in gradle and so on)
2. place the realease Image in generated/realeaseImage.jpg
3. edit the changlog.md (The release notes you would normally write on GitHub)
4. edit the releaseName.txt (single line only)
5. Commit everything
6. Create a new Tag and push it 
7. the workflow will create a Release with your release notes append the image and the generate the automated release notes upload the assets and start a discussion in announcements.


Donnnno:
WOW! Thanks for this setup, amazing really ✨
Gonna take a better look at it tomorrow hopefully!

PR #2225, opened 2024-07-30
adding: plattino, bestsecret, abrp

conny3496:


Donnnno:
Hi! Please use the template files to start with next time, your icons are huge compared to the others.

I've also cleaned them up: 

Scherm­afbeelding 2024-08-02 om 11 27 28

Donnnno:
Also make sure to set your appfilter to CF instead of CRLF

conny3496:
Thank you so much for your feedback and the corrections!

Issue #2227, opened 2024-07-31
[IDEA] Webp icons?


Donnnno:
Given the fact that Arcticons is getting pretty chonky, it might be a good idea to look into webp icons instead of pngs.
I know that Candybar supports it, I don't know how good launcher support is tho.

Investigating this at a point sometime later

Donnnno:
<img width="349" alt="Scherm­afbeelding 2024-08-02 om 13 16 20" src="https://github.com/user-attachments/assets/836c9eb1-25e8-4871-ac11-24a123fb7139">

<img width="361" alt="Scherm­afbeelding 2024-08-02 om 13 50 37" src="https://github.com/user-attachments/assets/3f54e454-a787-4b35-b29e-a4af3be98039">

Kaiserdragon2:
If the launcher support is there i could rewrite the script to convert to webp

kaanelloed:
I don't think it's a matter of launcher support. Unless, for some reason, there's a launcher that decompiles icon pack apk, they're all supposed to get the drawable from Resources.getDrawable(), so they'll get a BitmapDrawable whether it's a png or a webp. Since webp has been [fully supported](https://developer.android.com/studio/write/convert-webp) in Android since sdk 18, there should be no problem for any launcher.

Donnnno:
Great to hear that!

PR #2228, opened 2024-08-01
Contributer Image webp and Reproducible Build Fix

Kaiserdragon2:
image
With this these two blue runconfigs are also added.
RunChecks: runs the preparerealease.py --checkonly 
PrepareRealease: runs the preparerealease.py and updates the contributer images

With this there is no need to run the preparerelease.py script from the console 
The output from the script is printed in the run and build tab of android studio

#2227


Donnnno:
wooooooow that's sick!

Donnnno:
Noticed a small thing, the script can't count:

image

Kaiserdragon2:
Ahh i need to a config for a new release or icons will just be added as new to the old one.

PR #2241, opened 2024-08-06
Square and circle icons for the 'münster:app'

defy-all:
These are the square and circle icons for an app by the municipal utilities of the city of Münster in Germany.
Optimized SVGs right out of Inkscape.
Original icon can be found here: https://arcticons.com/extracted_png/munster_app.png

Please ignore the commit message on appfilter.xml. It's basically the same as this one.

If there are any problems with these icons please let me know and I will change them as suggested.


Donnnno:
Hi! looks like you deleted the whole appfilter file, please don't do that, and only add your lines of code to it in the future.
I've added the icon (square) to the app in a separate commit.

defy-all:
Hi, @Donnnno,

sorry for the inconvenience.

Don't know how that happened exactly and didn't realize it until after opening the PR.

Thanks for fixing this.

PR #2263, opened 2024-08-20
32 new icons

PurpleBabar:
Hey :) 

Here are new icons and attaching 2 more apps to one icon (running_h), and flux news is attached to books.svg.

Cheerz :)


Donnnno:
Made some changes with the letter based icons, other than that, great work!!
image

PR #2278, opened 2024-09-18
Added 1 more icon & created some icon mappings

sbstratos79:
Had to make yet another PR because I installed some new apps that needed icons.
New icon:
- LookYourLoot

Icon mappings:
- huggingAssist
- italki
- PhonePe


Donnnno:
Great!

PR #2282, opened 2024-09-23
6 new icons

kaanelloed:
It's been a while !

Here are 6 new icons


Donnnno:
It's been a while indeed, thank you for your contribution!

PR #2284, opened 2024-09-24
Update appfilter.xml

Donnnno:


Donnnno:
icons for 
- openbank
- fujifilm xapp
- panels
- immich
image

PR #2291, opened 2024-09-26
two radio icons

aloxe:
radio france new app
česky rozhlas


Donnnno:
nice!

alix-guillard-deltatre:
> nice!

thanks but I'd like to avoid you to redesign them, what would be the guideline I should follow?

aloxe:
> nice!

Thanks but I'd like to avoid you to redesign them. What would be the guideline I should follow?

PR #2312, opened 2024-10-05
Update Total Commander Wifi Transfer's icon

wwwwwwari:
This PR updates Total Commander Wifi Transfer's icon to match the looks of the main app's icon (totalcommander.svg). Not sure if the old version should be kept in this case but I've moved it to an alt.


Donnnno:
Nice!

PR #2326, opened 2024-10-14
Appfilter update

Argonlifesee:
Added support for:
- Symphony
- Password Store
- Seeker
- syncthing
- tor browser
- k9 mail
-  jellyfin
-  wetter


Donnnno:
Heya, found some issues!
What happened to these icons?
image
Also it looks like they're duplicates, as many of these already have existing icons that don't need an update

PR #2345, opened 2024-11-16
Update appfilter.xml

Kaiserdragon2:


Donnnno:
OMg thank you so much for adding these, was planning on doing it today, but you were ahead of me!

PR #2351, opened 2024-11-21
Icons for OSS Card Wallet and Light Meter

defy-all:
Last time I contributed, there was a problem with appfilter.xml.

I now found out why:
When clicking on 'edit file' in the browser Github shows an empty file. As I only pasted the info I gathered, the rest was deleted.

I now copied the contents of the file and pasted them along with the info on the two icons I created. I hope this works.

These icons were added:
Light_Meter
OSS_Card_Wallet


defy-all:
I fear there might still be a problem with appfilter.xml.

Maskedman99:
Try editing using Github codespace.

PR #2358, opened 2024-11-23
Dependency info block

### LINKED ISSUE #2357: Remove DependencyInfoBlock for F-Droid
We find that there is a `DependencyInfoBlock` in your APK.

It's a [Signing block](https://bi-zone.medium.com/easter-egg-in-apk-files-what-is-frosting-f356aa9f4d1) added by AGP and encrypted with the Google public key so it can't be read by anyone else except Google. You can read more about it [here](https://gitlab.com/fdroid/admin/-/issues/367) and [here](https://gitlab.com/fdroid/fdroidserver/-/issues/1056).

While this was added a while ago, we were only enforcing it for new apps, and recently we started scanning updates too.

Could you please disable it with the following code?

```
android {
    dependenciesInfo {
        // Disables dependency metadata when building APKs.
        includeInApk = false
        // Disables dependency metadata when building Android App Bundles.
        includeInBundle = false
    }
}
```

Thanks!

Kaiserdragon2:
Closes #2357

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -134,6 +134,13 @@ android {
         checkReleaseBuilds false
         lintConfig = file("lint.xml")
     }
+
+    dependenciesInfo {
+        // Disables dependency metadata when building APKs.
+        includeInApk = false
+        // Disables dependency metadata when building Android App Bundles.
+        includeInBundle = false
+    }
 }
 
 dependencies {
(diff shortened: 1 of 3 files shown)


Donnnno:
Thanks for adding this @Kaiserdragon2 !

PR #2360, opened 2024-11-23
RequestSite webp icons & Performance improvment

Kaiserdragon2:
Have the Request Site use Webp instead of png to reduce storage space needed
- [x] request.js
- [x] updatable.js
- [x] email_parser.py
 More?

Improved filterfunction in js to improve site loading with huge datasets

added funtion to email parser script that demotes (reduces request count) even if they are over the discard limit but aren't requested for some time


Donnnno:
ooooo awesome!

PR #2379, opened 2024-12-04
Icons for Revenge Manager and SEAtS

orbitaldr0p:
Yes I know Revenge's icon is kinda garbage but that's what the official icon looks like sooooo


Donnnno:
Looks fine to me :)

PR #2389, opened 2024-12-09
Rename an icon and add new icons from requests

gewaleelek:
* 5f18b0c: rename totalcmd_ftp to totalcmd_sftp because it is for the SFTP app, and add the icon for the FTP app
* d618076: add 10 icons from requests


Donnnno:
Awesome work!

PR #2390, opened 2024-12-10
Create drawable.xml at preparestep

Kaiserdragon2:
The drawable.xml was only created at build time so it wasn't updated.


Donnnno:
Ah, thanks for fixing this!

PR #2396, opened 2024-12-10
1 new icon

savetier:
- Stratum Authenticator


Donnnno:
No need for an icon when it's a rebrand! Thanks for the effor though :)

PR #2411, opened 2024-12-19
Categories

Kaiserdragon2:
adds logic for Category System
Counts Icons and adds it to Candybar

 #1938


Donnnno:
Fantastic work!

PR #2419, opened 2024-12-23
Change dependency for Play

Kaiserdragon2:


Donnnno:
From https://github.com/zixpo/candybar-sample/blob/main/app/src/main/res/values/dashboard_configurations.xml :  

```
    <!-- Enable/disable in-app review -->
    <bool name="enable_in_app_review">true</bool>
    <!-- After how many seconds the dashboard should count that the user visited the app.
        By default it is set to 30 seconds. If the user opens the app and stays for 30 seconds,
        it would be counted as a visit -->
    <integer name="in_app_review_visit_time">30</integer>
    <!-- After how many visits the in app review dialog should pop up -->
    <integer name="in_app_review_after_visits">2</integer>
```

Kaiserdragon2:
i have set it to 10 sec instead of the 30 and the visits is sill 2

Donnnno:
Thanks for adding this!
The Gradle setup looks super wonky, but that's just the trait off with all the versions we have 😅 
I'm curious to see if this boosts the reviews (:

PR #2433, opened 2024-12-27
Update `eu.novapost` component name

nazar-pc:
Apparently not an example anymore :see_no_evil:


Kaiserdragon2:
instead of changing the component of an existing entries it's good practice to just add it.
So people with an older version still have the icon applied.

nazar-pc:
I can do that, but I'd argue it makes the most sense for applications that can work offline.

This is a frontend to a delivery service though 100% dependent on servers. With application auto-updates being the default (and strongly recommended for security reasons) as well as older versions likely to have issues or not work at all as backend evolves over time, the value of occupying space in the file is very questionable.

Let me know if you still prefer to add instead of update these.

PR #2440, opened 2024-12-28
fix missing end tag

Kaiserdragon2:


Kaiserdragon2:
Just created something that will check for parsing errors will implement it probably tomorrow.

PR #2441, opened 2024-12-28
3 new icons

Random-Penguin:
Added:
- Breasy
- CIC
- CIC Pay

Please check cic_pay.svg because I believe I implemented the font poorly. I'd love to hear what I'm doing wrong and how I can fix it.


Donnnno:
Hey hey, congrats on your first contribution!
The icons look good, only the font needed to be corrected, like you mentioned already. We use the SVG version of the font, located inside the template folder. 
Other than that, the appfilter doesn't need a separate code block, just put it in the resources bit between all other lines next time :)

image

image

Random-Penguin:
Just real quick, do you know if/how the SVG version of the font can be imported to Vector Asset Creator?

PR #2446, opened 2024-12-30
Add 100 icons

wwwwwwari:
This PR adds:
* Appfilter updates for 10 existing icons
* 100 icons (separated to 5 commits, 20 each)

I had to set up InkScape again so if there's any weird things in my icons just let me know 🙂

Happy holidays!


Donnnno:
🤯

PR #2450, opened 2024-12-31
Allow Automated Dashboard Updates

Kaiserdragon2:
For this to work you need to update the Environment secrets in [Request-Update](https://github.com/Arcticons-Team/Arcticons/settings/environments/4778934031/edit)
image


Donnnno:
I'm currently looking at Infomaniak's email service. They respect privacy and offer free/very cheap tiers with custom domain possibility.  

https://www.infomaniak.com/en/ksuite/service-mail
https://www.infomaniak.com/en/support/faq/2427/sync-emails-across-all-your-devices

So if all is configured correctly that should be a suitable option instead of my private proton mail space :)

Donnnno:
@Kaiserdragon2 I'm not sure what to do with the merge error for main.yml, could you take a look at it when you have time?

Kaiserdragon2:
You don't need to merge this branch anymore i just kept it open as a sort of reminder. 
At the time i have rewritten the workflow i implemented this to activate it the variable for automatic updates needs to be set to true. See the last comment.

Donnnno:
Aaaah of course, thanks! 😵‍💫

Donnnno:
Currently, does it fetch all mails from a mailbox or only request ones?
I'm asking because in an ideal scenario I have two mails: hello@arcticons.com for general questions and requests@arcticons.com. How it's set up right now is that both go to one mailbox, but in different folders there.

Kaiserdragon2:
I fetches all Mail from the inbox and processes all unread mail and ten matches the subject. If the subject matches it gets processed further and set to read.

Donnnno:
Ah yeah, great!

Donnnno:
I've entered all secrets now, and enabled the variable!
Let's hope it works correctly :)

PR #2453, opened 2024-12-31
Update games.xml

wwwwwwari:
This PR updates games.xml. It also sorts the file (because it hurt my elderly eyes trying to check whether a game already exists in an unsorted file...)


Donnnno:
sick!

PR #2466, opened 2025-01-01
Automated Release Image and Name Generation

Kaiserdragon2:
This generates the release Image from all new Icons 
and the Release Name doesn't need to be written into the txt file.


Donnnno:
wow, we're going fully automated this year!

Donnnno:
I'm curious, do you have an example on how it looks?

Kaiserdragon2:
For the release name the workflow grabs the GitHub tag name and appends "Release".

For the image every time prepare Release is run  this checks which icons are i newdrawables.xml and how many then it calculates how to lay them out in a square grid. Then the svgs are combined to a collage.

The webp image with this pull request is generated with it. 

You could also run the image generation by opening Android studio and in the preparereleasehelper section run the imagecollage file.

Donnnno:
Cool, will try later!

PR #2478, opened 2025-01-06
Fix path detection in preparerelease.py

### LINKED ISSUE #2477: preparerelease.py does not detect the Arcticons directory properly
`check_arcticons_path` in `preparerelease.py` is looking for the old "other" directory instead of "newicons".

It thus displays the warning `The path 'path/to/arcticons" does not include the 'Arcticons' folder` when invoking the script properly, i.e. with: `python scripts/preparerelease.py path/to/arcticons`.

rdehenne:
Closes #2477

### DIFF
--- a/scripts/preparerelease.py
+++ b/scripts/preparerelease.py
@@ -21,8 +21,8 @@ def check_arcticons_path(path):
         return arcticons_folder
     else:
         app_folder = os.path.join(path, "app")
-        other_folder = os.path.join(path, "other")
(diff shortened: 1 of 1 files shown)


Kaiserdragon2:
Nice that looks good. 
Just as a note if you want to build Arcticons yourself and thus run this script, you should instead use the run configurations in Android studio

rdehenne:
Yeah, it took me some time to figure out how to of build Arcticons / Snow, but I found out there simply was an Android Studio configuration in the end. Thanks for the advice though! 😊

PR #2484, opened 2025-01-10
Everything from the pythonscript is now in Java

Kaiserdragon2:


Donnnno:
Damn, that's some advanced scripting! Awesome work!

PR #2487, opened 2025-01-11
Add and update icons

gewaleelek:
* Add new mappings for fossify draw and racing games in e0851e4
* Update red reader alpha icon to match the letter_uppercase_square_r design used for the normal red reader app in 8a5d4c6
* Add icons from requests, mostly games in 2d99450

My game icons still look complicated, feedbacks are welcome :slightly_smiling_face:


Donnnno:
Amazing work on these!
The only thing I changed is the talking tom 2 icon: removed the 2 because the icons are already different enough from eachother

image

PR #2498, opened 2025-01-16
Fix: icon request links are the wrong color

### LINKED ISSUE #2493: Links on request section in app are unreadable
Screenshot_20250114-235153

kaanelloed:
Fix #2493

The links in candybar are based on the accent color, so here's a fix for each TextView. I haven't found a way to apply this more globally.


Donnnno:
Thanks for the fix! Good to know :)

PR #2500, opened 2025-01-17
Fixes

Kaiserdragon2:
properly calculate image size (Release image)
Sort system and games file
allow icons in multiple categories
have new release overwrite newDrawables.xml


Donnnno:
niceee

PR #2502, opened 2025-01-19
New email address 

Donnnno:
Linked to #2450 which allows us to fully automate the requests process. 
We've changed mail provides from Proton to Disroot because it allows for the use of IMAP configs. 
Since emails are no longer e2e-encrypted through Proton's servers, I've changed the privacy policy too.


Donnnno:
Sadly still haven't got a reply from Disroot, so I've configured everything through proton for now.

PR #2542, opened 2025-01-27
Kanji and Kana alterations for consistency

HaleyHalcyon:
kanji
This commit also adjust the alignment of the letters in the UNIQLO logo to make it look more even.


Donnnno:
Nice work on these!

PR #2543, opened 2025-01-27
New icon iubilaeum25

desk7:
Hi,
that is my first contribution here, and I hope I did a good job. The icon is quite simple but I couldn't find the original svg so I use the raster image as starting point.


Donnnno:
looks good

PR #2548, opened 2025-01-28
icons for 3 apps

Aneeq-Tahir:
added 3 icons for burrito bison (game) and hadaf & PGC (both with the same icon) but added their icons separately


Donnnno:
Please don't delete the whole appfilter when adding entries! I'll have to add your icons seperatly now :s

PR #2552, opened 2025-01-30
Update Appfilter

loxoron218:
"Digitales Bezahlen" is a german app that allows paying via NFC. It has a very similar icon to the default "Nfc tools", hence a new one isn't needed.


Donnnno:
thank you fro this addition. In order for it to work, you'll need to change the drawable name from digitales_bezahlen to nfctools!

loxoron218:
I just updated it! :)

Donnnno:
Thank you!

PR #2554, opened 2025-02-02
Add 3 new icons: inter_profile_sharing, ozlo and dreisechzehn

je223xa:
- app dreisechzen does not fully follow the guidelines as the style of the icon uses 0.5 thin letters at the bottom. feel free to remove it. you can look at the original icon [here]( https://i0.wp.com/dreisechzehn.de/wp-content/uploads/2020/03/316_Logo-Vertical-White-BG_BlackCircle.png)


Donnnno:
Found two reference images still inside the icons, and lines having inconsistent widths, so fixed that :)

PR #2564, opened 2025-02-04
Add Neko Atsume 2 and Crowdin icons, and a minor change

wwwwwwari:
This PR adds icons for Neko Atsume 2 and Crowdin (the latter has no appfilter). It also groups the appfilter entries of Pant together (the upper one might actually be a typo, but I'm not familiar with the app).


Donnnno:
Neko Atsume my beloved

PR #2571, opened 2025-02-05
23 icons and mapping updates

gewaleelek:
* 2d779e9: update mappings, most of them are from https://arcticons.com/updatable.html
* 9803fe9: add 23 icons


Donnnno:
Wow, these must have been a lot of work!
Great job making these :)

PR #2573, opened 2025-02-05
Pages updates

Donnnno:
we're in 2025 now!

### DIFF
--- a/docs/css/arcticons.css
+++ b/docs/css/arcticons.css
@@ -89,11 +89,6 @@ section {
   color: #fff;
 }
 
-.copyleft {
-  display: inline-block;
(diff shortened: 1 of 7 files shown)


Kaiserdragon2:
@Donnnno You deleted the js in the request.html while adding the footer this kills the dashboard.

Donnnno:
oh oops, that wasn't my intention

Donnnno:
removed the footer from that page because it was overlapping with the categories, anyways it's not necessary.

Kaiserdragon2:
> oh oops, that wasn't my intention

Thought so, looked like accidentally marking a line while copy pasting.

PR #2619, opened 2025-02-17
Update Appfilter

loxoron218:
The app "Beer With Me" doesn't use the proper icon anymore, since it changed for whatever reason its name to:
`se.dagsappar.beer/se.dagsappar.beer.MainActivity`

Still, I left the old app name on the _appfilter_ list, in case some people are still using the old version. If you prefer to keep the list as simple as possible, I can of course delete the old name :)

PS: and deleted an unnecessary line on the second commit


Donnnno:
looks good!

PR #2621, opened 2025-02-17
Some emoji

Kaiserdragon2:


Donnnno:
Out of curiosity, are the emojis based on existing ones or are they original creations? :)

Kaiserdragon2:
The are based on existing ones

Donnnno:
ah which set was it again, might wanna add some too!

Kaiserdragon2:
Openmoji

PR #2623, opened 2025-02-18
add logline and chooser icon

UrAvgCode:
adds an icon for the logline movie app.
https://github.com/Patch4Code/Logline

adds an icon for the chooser app.
https://github.com/UrAvgCode/Chooser


Donnnno:
looks good!

PR #2632, opened 2025-03-02
Update Alembicons

kaanelloed:
I will change the namespace of Alembicons in the next release.

Also, use the Alembic emoji as the app icon.


Donnnno:
Sounds good!

PR #2637, opened 2025-03-10
Rename some icon files

Npepperlinux:
Unified the file name format by changing some alternative icon file names from `_alt[0-9]` to `_alt_[0-9]` and adding underscores to old GooglePlayStore Alt icon file names.


Donnnno:
Thank you for these fixes!

PR #2651, opened 2025-03-19
Upgrade ebtart, add bagatur chess, cartes IGN and MobiTech metros

aloxe:


aloxe:
There are many MobiTech apps with metro maps in many cities. They all have the same icon with the name of the city. I chose to name it mobitech and let all apps us the same file.

Donnnno:
Nice work! Did some resizing, and made fixes to the line width.
I've also removed the Ebtart icon in favor of the already existing currency_euro one :)

PR #2659, opened 2025-03-24
add cashinator icon

UrAvgCode:
adds an icon for the cashinator app.
https://play.google.com/store/apps/details?id=net.cashinator.www&hl=en


Donnnno:
Nice!

PR #2666, opened 2025-04-03
ReWASD Icon

phizlip:
Icon for ReWASD, a controller remapping utility


Donnnno:
Congrats on your first contribution!

The icon looks good! I've resized it a little bit, and fixed the stroke width and corners. 

I also found that the reference image was still inside the SVG, so make sure to remove it from your artboard before exporting ;) 

image

PR #2667, opened 2025-04-03
Strike Icon

phizlip:
Icon for Strike, a Bitcoin exchange


Donnnno:
Same feedback as the previous one, but I also noticed a huge amount of anchor points, so I cleaned the shape up:

Also feel free to put all your icons in one PR next in the future :) 

image

PR #2669, opened 2025-04-03
Add various icons

phizlip:
Adds icons for the following apps:

[Briar Mailbox ](https://play.google.com/store/apps/details?id=org.briarproject.mailbox)

[ToffeeShare](https://play.google.com/store/apps/details?id=com.toffeeshare.android)

[Qlango](https://play.google.com/store/apps/details?id=co.uk.exocron.android.qlango)


Donnnno:
Found some errors in the icons:
Brair Mailbox had expanded strokes (so fill elements instead of strokes), Also squashed corners (not really a fan of those)
qlango & toffeeshare had many unnecessary anchor points, so fixed that too.
image

image


image

PR #2670, opened 2025-04-05
Updated appfilter for ZDF, Quillpad

KR4U7H3R:
I never delete old entries, as I do not know if it will break thinks for others.


Donnnno:
yeah, it should be fine to leave the old ones in :)

Issue #2694, opened 2025-04-11
Build preparehelper failed


Donnnno:
<img width="918" alt="Image" src="https://github.com/user-attachments/assets/c8f5259c-d352-4b8e-98ab-42f2404218e1" />

```
Exception in thread "main" java.lang.UnsatisfiedLinkError: /private/var/folders/vk/7v0bz8xs2vd15g592nvhk6_h0000gn/T/15763771451007316559libwebp-imageio.dylib: dlopen(/private/var/folders/vk/7v0bz8xs2vd15g592nvhk6_h0000gn/T/15763771451007316559libwebp-imageio.dylib, 0x0001): tried: '/private/var/folders/vk/7v0bz8xs2vd15g592nvhk6_h0000gn/T/15763771451007316559libwebp-imageio.dylib' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e' or 'arm64')), '/System/Volumes/Preboot/Cryptexes/OS/private/var/folders/vk/7v0bz8xs2vd15g592nvhk6_h0000gn/T/15763771451007316559libwebp-imageio.dylib' (no such file), '/private/var/folders/vk/7v0bz8xs2vd15g592nvhk6_h0000gn/T/15763771451007316559libwebp-imageio.dylib' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e' or 'arm64'))
	at java.base/jdk.internal.loader.NativeLibraries.load(Native Method)
	at java.base/jdk.internal.loader.NativeLibraries$NativeLibraryImpl.open(Unknown Source)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(Unknown Source)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(Unknown Source)
	at java.base/java.lang.ClassLoader.loadLibrary(Unknown Source)
	at java.base/java.lang.Runtime.load0(Unknown Source)
	at java.base/java.lang.System.load(Unknown Source)
	at com.luciad.imageio.webp.NativeLibraryUtils.loadFromJar(WebP.java:131)
	at com.luciad.imageio.webp.WebP.loadNativeLibrary(WebP.java:31)
	at com.luciad.imageio.webp.WebP.<clinit>(WebP.java:38)
	at com.luciad.imageio.webp.WebPEncoderOptions.<clinit>(WebPEncoderOptions.java:20)
[...]

PR #2701, opened 2025-04-13
Alternative approach for Mac arm64 issue

Kaiserdragon2:
#2694
Could you @Donnnno test this one a Mac it uses a fork of the imageIo dependency that has Mac support.


Donnnno:
oh that would be much better than the mess I've created 😅

Donnnno:
Seems like this solved the issue!
There are two red lines when I run it, I don't know how relevant it is because it works fine:

Scherm­afbeelding 2025-04-14 om 09 32 22

Donnnno (approved):

PR #2711, opened 2025-04-16
Smart Launcher patch

Donnnno:
A patch making sure that material you icons work with the new Smart Launcher function that themes non-covered icons based on the color that we're using.


Donnnno:
#2741

PR #2718, opened 2025-04-19
Introducing Universal Termux Icons

leoclrks:
I created those icons because I want to differentiate other Termux applications.
It's created because I use Kvaesitso and I disabled app labels, so with one icon, it's hard to choose between applications.
tempFileForShare_20250419-134328
_Pic 1. Confused Termux icons._
So I created eight different Termux icons for different apps, like Float, API, Boot,...
And you see in the pics, it's easy to know.
Termux:Styling isn't appeared in app drawer, but I created one for it.
Screenshot_20250419_134024
_Pic 2. One Termux Icons and Universal Termux Icons._
_P/s: I don't want to be credited._


Donnnno:
Nice one, thanks for creating these!

PR #2725, opened 2025-04-19
New icon for huafetcher based on smartwatch.svg

khwolf:


Donnnno:
Hey! if you got lots of commits like these, it's also fine to bundle them into one :)

khwolf:
OK, will do. Thought it would be easier to reject on an individual basis.

Donnnno:
Thanks, it's easier for me to look at everything at once instead of merging everything loose. I'm opening all svgs in Illustrator to check them :-)

PR #2733, opened 2025-04-23
Add existing icons to some matching apps

khwolf:


Donnnno:
Nice, thank you for adding these!
One small style thingy: make sure to use tabs instead of spaces, that way the appfilter will look more readable ;)

image

khwolf:
> Nice, thank you for adding these! One small style thingy: make sure to use tabs instead of spaces, that way the appfilter will look more readable ;)

OK. Almost every other project does not want tabs, but i will do my best to fulfil your request.

PR #2735, opened 2025-04-24
Icon for MiCTS, Delta Force, and Waifu2x ncnn

kato1101:


Donnnno:
Great work on your first commit! Just minified the svgs :)

PR #2750, opened 2025-04-29
Add 4 new tabletop-related icons

troxor:
Here is a handful of icons related to tabletop games
- Infinity Army
- Comlog
- Bushido Companion
- paintRack


troxor:
As an aside, I attempted to request and then fulfill the request myself, involving sending from a non-default e-mail address. While I'm not sure if I did anything wrong, there doesn't seem to be any issue just contributing without an official request; the instructions were pretty straightforward!

Donnnno:
Hey, nice work on the icons!

> I'm not sure if I did anything wrong

No, it's just very unlikely with the queue that we have at the moments that icons will get made. https://arcticons.com/requests.html So, very cool that you decided to make them yourself instead :)

PR #2752, opened 2025-04-30
Update appfilter.xml

voxelized-voxel:
Added support for Pipebender and Pipebender Extended via existing icon

Closes https://github.com/Arcticons-Team/Arcticons/issues/2737.


Donnnno:
Hey, looks like you found out how to add it yourself, congrats!
Two things I noticed while checking your PR:
- Seems like the appfilter was converted to CRLF instead of LF, this is not huge, but GitHub detects it as modifying every line, so make sure it's LF when you commit it. (can be easily changed in the bottom with VS Code)
- We place new lines in the appfilter in the `/newicons/appfilter.xml` folder, instead of the one inside the app directory.

PR #2753, opened 2025-04-30
New icons for apps

khwolf:
- EVNotify
- CBTAndroid
- Imprint
- Meeting Owl

(all in white and with tabs) ;-)


Donnnno:
Nice work!

PR #2765, opened 2025-05-11
Update Cloudstream icon and add IED

JFronny:
The IED icon is kind of ugly, but that is also true for the [normal icon](https://codeberg.org/Krixec/IED-FDroid/src/branch/dev/metadata/en-US/images/icon.png)


Donnnno:
Nice!

PR #2779, opened 2025-05-16
Four New Apps

phillip-bruno:
- Adds "Amazon Workspaces", which reuses the existing `amazon_aws_console` icon
- Adds "iCSee", a security camera app
- Adds icon for the "Sesterce" expense tracking application.
- Maps the "Booming Music" offline music player to the music icon.

Hopefully my additions are okay, please let me know if I need to revise anything!


Donnnno:
Thanks for the icons!

I noticed that the Sesterce icon is just a dollar symbol, so I've replaced it with the already existing dollar icon.

PR #2781, opened 2025-05-17
Add 6 icons

Crazy-Unicooorn:
- Famileo
- Find Family
- Evil Spin
- Jigsaw
- Whisper
- Omni

image


Crazy-Unicooorn:
Question, will it remove automatically from https://arcticons.com/requests after merging?

Donnnno:
Great work!
Removed the whisper icon, as it doubles with other recorder icons (used one of those instead).
I've also done some resizing for consistency

image

PR #2786, opened 2025-05-25
update appfilter.xml

WJehee:
Updated appfilter.xml for NS app


Donnnno:
No more bright yellow icon! 😄

PR #2787, opened 2025-05-25
Added pluxee fr icon

lyghtnox:
Hey,

Updated appfilter to add the Pluxee FR icon.

I also renamed `Pluxee España` to `Pluxee` as there are multiple countries in the category, each with their own app but a similar logo.


Donnnno:
Ah thanks for the addition!

PR #2795, opened 2025-05-31
Add files via upload

Bromerz:
Alternative icon for Man Utd app. Man_Utd_Alt


Donnnno:
did a bit of shape cleanup here, it was a bit too clutered

PR #2798, opened 2025-06-03
minimiszed style elements for Linux

Npepperlinux:
Related to https://github.com/Arcticons-Team/Arcticons-Linux/issues/92
Fixed some icon CSS style elements manualy so that scripts used in Arcticons-Linux are not confused where stroke-width should be changed.
(Applied directly under `icons/(balck|white)` as it does not affect the appearance of WEBP for Android.)


Donnnno:
Nice!

PR #2793, opened 2025-05-28
Update appfilter.xml

Npepperlinux:
- 3D Earth Pro (use "Earth" icon)
- some camera apps
- DJI Fly
- John GBA & GBC (use "gbaemu" & "gbcc" icon)
- Metronome
- MyMTN
- PC Creator PRO
- Phone Manager
- PPSSPP Gold
- Stick War: Legacy
- Talking Tom Time Rush (use "Talking Tom Gold Run" icon)
- That Shopping List
- some widget apps
- Yodobashi Gold (use "Yodobashi" icon)

PR #2802, opened 2025-06-05
Add 7 icons

fuchse-ohren:
- ATOK for Android
- ATOK for Android[Professional]
- dアニメストア
- ENEOS公式アプリ
- ヨドバシゴールドポイントカード
- Moneytree
- Nature Home


Npepperlinux:
Hello @fuchse-ohren !
I have a few suggestions for your first pull request.

1, Arcticons already has an [ATOK icon](https://github.com/Arcticons-Team/Arcticons/blob/cea6d144a9526a9c6e4973fb158c7400042c1f6c/icons/white/atok.svg), so there is no need to add `atok.svg` again. Adding the string "PRO" to the already existing ATOK icon and adding `atok_pro.svg` is a good suggestion. 
2, "dアニメストア" should be translated as "d Anime Store" and renamed from `d.svg` to `d_anime_store.svg` with `drawable="d_anime_store"`.
3, "ヨドバシゴールドポイントカード(GoldPoint)" is set to use the "Yodobashi" icon in #2793, so the next release will show "Yodobashi" Arcticons. It may not be necessary to add `goldpoint.svg`. (If you need to distinguish between "Yodobashi" and "Yodobashi GoldPoint", we may consider adding `yodobashi_goldpoint.svg` with "G" added to the Yodobashi icon)

If you have time, consider adding a new commit to this pull request's branch based on these suggestions.

fuchse-ohren:
Thank you for the suggested revisions!

1,3. I was referring to the current release version and did not track changes prior to the release. Thank you, I would like to withdraw my request for ATOK and GoldPoint.

2. I have adopted the naming convention as it is in the "Icon Request" app outputs. I accept the suggestion that it should be changed to avoid duplication with other apps, and I would like to modify it as well.

PR #2811, opened 2025-06-09
Delete duplication in `/newicons`

Npepperlinux:
Looks like there was some kind of mistake, one more "MTS Cloud" and one more "BuzzFeed" icon.


Donnnno:
Oof, my bad! Thanks for fixing it :)

PR #2821, opened 2025-06-17
Add Syncara app icon

constellationz:
Added app icon:

- Syncara

P.S. I switched to Android about a week ago and I found this project on Accrescent. Glad to be here!


Donnnno:
Heya, sorry for the late reply!
Great work on the icon :)

PR #2823, opened 2025-06-17
newiconsbyjustxbird0.1

justxbird:


Donnnno:
Great work on the the icons, I've made some tweaks and I had to recreate one of em.
Noticed that one icon had outlined shapes, so had to fix that:
image

PR #2833, opened 2025-06-25
add new icons for GlideX and Unibuddy

Thrasher146:


Donnnno:
Heya, nice work on the icons!

PR #2845, opened 2025-07-01
added 3 icons and fixed 1 icon

camradex:
added 
- Good Coffee, Great Coffee 
- mo.co 
- Easy Notes 

fixed a mistake on the "freecell" icon i previously contributed


Donnnno:
thanks for the icons!

PR #2854, opened 2025-07-09
Update Canta icon

5Litt:
As of update 3.0.0, Canta has switched its package name, with the older
one being deprecated. This commit adds the corresponding icon to the new
package name. The old entry has been left untouched for backwards
compatibility.


Donnnno:
Thanks for the addition!

PR #2858, opened 2025-07-11
Add My Max Fitness icon

5Litt:
Adds an icon to the package `com.datlink.mymaxfitness`.
The svg file is already optimized.


Donnnno:
Nice!

PR #2865, opened 2025-07-14
Add icon for baba is you

EbbDrop:
I have simplified the icon a lot to just use baba without the text or keke:
baba_is_you
Screenshot_2025-07-14_22-42-05
In my opinion it is still very recognizable and unmistakably _baba is you_, but it is a lot cleaner.


Donnnno:
Omg Baba is you!

Added Keke as an alt 😄 

Scherm­afbeelding 2025-07-17 om 16 40 30

EbbDrop:
Nice! thanks for reviewing

PR #2891, opened 2025-07-28
Upload new icon for myTrip

Kazuala:
Uploaded a .svg for the app myTrip and edited the appfilter.xml with the correct information.

### DIFF
--- a/newicons/myTripIcon.svg
+++ b/newicons/myTripIcon.svg
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<svg
+	version="1.1"
+	xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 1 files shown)


warimiya:
Hi! Just helping with some quick check since I'm on my phone:
- The icon's file name has to match the drawable name in the appfilter.xml. Since it's `drawable="mytrip"` in that file, the icon should be named mytrip.svg (all lowercase)
-  From the svg, it seems the stroke color is #fdfdfd, not #ffffff?
    - ETA: also stroke width seems to be 0.96 instead of 1 if I'm reading the svg correctly
- You might want to credit yourself following [this guide](https://docs.arcticons.com/contribute/credit-yourself) 😄

PR #2893, opened 2025-07-30
new icon

user12w:
added find my ip icon


Donnnno:
Hi, thanks for your contribution.
Had to fix a couple of things, I redid the icon to make sure everything was a line-based vector instead of a type element.
You also created a text file instead of inserting your line into the main appfilter.

PR #2896, opened 2025-08-01
New Icons for Solarman Zendure and Smartbroker

Winux-Arch:
Added new icons for solarman business, zendure and smartbroker.


Donnnno:
Nice work on the icons, and congrats on your first PR!

PR #2927, opened 2025-08-10
Adds mappings for five new apps.

phillip-bruno:
Hopefully a simple addition, I added mappings to four new apps to existing icons. With a new package name for the Dragonpass app after a redesign/refactor that it went through.

Let me know if you want anything changed!


Donnnno:
thank you! looks good :)

PR #2978, opened 2025-08-26
Add 15 icons

stratself:
Extra notes:

- Update owncloud to new icon and replace old as alt
- Switch man_utd main and alt icon
- Circle K icon used for 2 apps


Donnnno:
Awesome work on the icons!
I've tweaked some tiny things like:
- joining lines together when it was possible
- making sure every line is 1px, with rounded caps and corners
- removing a trademark from the circle_k icon

PR #2987, opened 2025-08-27
Add 9 icons for 10 apps

stratself:
- CGV_cinemas icon is for 4 apps 
- Vanity Forgejo icons as tribute to the software that Codeberg runs on


Donnnno:
great!

PR #2988, opened 2025-08-27
Add 3 icons

enizor:
Hi, first PR here, I'd love to have some feedback on these icons.

Original icons:

ANCV
Mes Comptes
EFS

I traced the Chèques-Vacances icon and ignored the text so it remains off-centered.  Should I re-center the bird?


Kaiserdragon2:
Yes i think it should be centered otherwise it would look weird beside other icons.

Donnnno:
Hi!

I've made some small modifications to your icons, here's a preview:

image

No big changes, but I noticed strokes with 75% opacity, and some double lines that could be reduced to a much simpler shape.
Nice work!  :)

enizor:
> No big changes, but I noticed strokes with 75% opacity, and some double lines that could be reduced to a much simpler shape. Nice work! :)

Sorry about that, I used partial opacity to check the shape and probably forgot to turn it back.
I don't know much about SVGs so the shapes certainly aren't optimized.

I also noticed you changed the "size" (whatever that means for SVGs) to 300px. I used 48px from the documentation - should I change my base template?

Thank you for your corrections :).

Donnnno:
Idk why it changes it to 300px, it's probably a github thing :s

PR #2989, opened 2025-08-28
Update appfilter.xml

loxoron218:
The letterboxd icon stopped working, probably because the app's name changed?


Donnnno:
omg thanks, noticed that too this morning 😄

PR #3003, opened 2025-09-06
Add 1 app and modify appfilter.xml

tapeykatt:
App added:
- Skylight

I also added an additional appfilter for ibisPaint X and Xed Editor because, after updating the apps, its icons returned to it's original ones.


Donnnno:
nice!

PR #3014, opened 2025-09-18
2 apps have changed drawable path

aloxe:


aloxe:
Is this correct if I add this change [here instead](https://codeberg.org/Arcticons/Arcticons-Android/pulls/3016)?

PR #3065, opened 2025-11-14
Add new 3 icons and update contributors/appfilter

RunTheBot:
Added SVG icons for Echo Nightly, InstaEclipse, and TDSB Connects. Updated appfilter.xml to include these new icons and registered their respective components. Added RunTheBot as a contributor in contributors.xml.


Donnnno:
Nice work! 
I've changed the tdsb icon to use the Arcticons font so that it fits with the rest :)

image

PR #3066, opened 2025-11-14
Fix stroke-width in newicons

Npepperlinux:
It seems something with a stroke width other than 1px was mixed in.


Donnnno:
Oh, thanks for checking!

PR #3106, opened 2025-11-27
Update appfilter.xml

Npepperlinux:
Added some entries and separated the Opera Mini entries from Opera.

Question: KernelSU Next entries keep increasing with unique IDs. Should we refrain from adding more, like Luckey Patcher?


Donnnno:
Hmm good question; that's a tricky one. I'm fine with adding more entries if it's easy for us to add.

PR #3112, opened 2025-11-30
New Contributor | New icons

viru185:
- Fix indentation for Veepee entry in appfilter.xml
- Add Viren Hirpara as a contributor in contributors.xml
- Add HDFC Bank entry to appfilter.xml
- Add Truthin icon to appfilter.xml and include SVG file


Donnnno:
Nice work on the icon, just scaled it a little bit to fit between the reset :)

viru185:
# can we have some guid?
 1. how big is the icon should be in the view box?
 
- there is guid for the icon size which is 48px X 48px. however, I was confuse when I design the `Truthin` icon. how big should I make it. 
1. should I make it fit the viewbox !?
2. should I make it fit the square provided in templates folder !?
3. or should I just eye boll it !?

- I did look up for the similar icons and compared the size but I could make any sense of the size.
 > It would be great to have the guide for `how to size the icon?` on the site.

PR #3140, opened 2025-12-17
Creating an icon for the ADT+ app

CallmeJoeBob:
Adding an icon for the ADT+ app


Donnnno:
Nice work on the icon!
I noticed that the reference image was still present (but hidden inside the svg), so I removed that from the file :)

PR #3161, opened 2025-12-25
New icons and updated app filter

Elaborendum:
Fixed some appfilter entries and added icons.

There are alternative icons for AMBOSS and Lecturio. Also icons for the obwheel.quartertone.net and play.pokemonshowdown.com webapps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 </div>
   
 # Arcticons for Android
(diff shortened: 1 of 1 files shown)


Kaiserdragon2:
i don't know what went wrong but something did. Instead of editing the current appfilter.xml you created a new one. Also there is a file named n that does not need to be there. I would recommend creating a new branch from upstream/main and add your changes there.

Elaborendum:
Sorry, the guide on the website mentioned something about appfilter being in the newicons folder, I modified the appfilter on other this time. Is it ok now or should I instead start over?

Elaborendum:
Forget it, I see the issue now I think. No idea how I missed the folder.

PR #3162, opened 2025-12-26
Added 24 icons, fixed 1 and edited appfilter

Elaborendum:
Fixed the real racing 2 icon.

There are alternative icons for AMBOSS and Lecturio. Also icons for the obwheel.quartertone.net and play.pokemonshowdown.com webapps.


Donnnno:
Nice work on the icons; I had to make some fixes here and there because I found some inconsistencies with line width, clipping masks, and small text elements.

PR #3163, opened 2025-12-26
WIP: Dashboard redesign

Donnnno:
Todo:

- [x] Fix the button scaling mess:
- [x] Highlight activated buttons (regex, categories, and matches) would allow hiding text in compact view
- [x] Fix Regex search
- [x] Fix app store buttons
- [x] Fix category search
- [x] Floating copy button
- [x] Actually center the floating copy button.
- [x] Hide the floating button when nothing is selected.
- [x] Escape button to close popups
- [x] Global theming for the tables
- [x] smaller font for header on small screens ?
- [x] add Download option for selected (floating)
- [x] Better theming for the toggles
- [x] Rework image preview to work with escape like the other popups. 
- [x] add keep names button to the rename overlay.
- [x] Updates page: add Arcticon for comparison to see if it matches. (Needs some JS to be perfect)

### DIFF
--- a/docs/dashboard/css/buttons.css
+++ b/docs/dashboard/css/buttons.css
@@ -93,6 +93,12 @@ background-color: hsl(from var(--primary) h s 50%);
   animation: 1s longpress;
 }
 
+.btn-row {
+    display: flex;
+    gap: 10px;
+    margin-top: 10px;
+}
+
 @-webkit-keyframes longpress {
 
   0%,
--- a/docs/dashboard/js/const.js
+++ b/docs/dashboard/js/const.js
@@ -75,5 +75,6 @@ export const DOM = {
     imagePreviewTitle: document.getElementById('preview-image-name'),
     floatingBtnContainer: document.getElementById('floating-action-btn'),
     downloadSelectedBtn: document.getElementById('download-selected-btn'),
-    closePreview: document.getElementById('close-preview')
+    closePreview: document.getElementById('close-preview'),
+    keepBtn: document.getElementById('keep-button')
 };
--- a/docs/dashboard/js/functions.js
+++ b/docs/dashboard/js/functions.js
@@ -70,10 +70,6 @@ export function CopyAppfilter(index, rename) {
     let copyText = "";
     const node = document.getElementById("drawableName-input");
 
-    // Handle rename mode
-    if (rename) {
-        document.getElementById("renamer-overlay").classList.remove("show");
-    }
     // Multi-row mode
     if (index === null) {
         copyText = getSelectedEntries()
--- a/docs/dashboard/requests/index.html
+++ b/docs/dashboard/requests/index.html
@@ -9,7 +9,7 @@
     <link rel="icon" type="image/webp" href="/img/arcticons/icon.webp" />
     <link rel="stylesheet" href="/dashboard/css/dashboard-main.css" />
     <link rel="stylesheet" href="/dashboard/css/dashboard-header.css" />
-    <link rel="stylesheet" href="/dashboard/css/dashboard-table.css" />    
+    <link rel="stylesheet" href="/dashboard/css/dashboard-table.css" />
     <link rel="stylesheet" href="css/dashboard-table-main.css" />
     <link rel="stylesheet" href="/dashboard/css/slider.css" />
     <link rel="stylesheet" href="/dashboard/css/dashboard-popup.css" />
@@ -164,7 +164,10 @@
                         title="Change drawable name  to ..." />
                 </div>
                 <hr class="separator">
-                <button class="btn" id="rename-button"> Rename drawable </button>
+                <div class="btn-row">
+                    <button class="btn" id="rename-button"> Rename drawable </button>
+                    <button class="btn" id="keep-button"> Keep Original Name </button>
+                </div>
             </div>
         </div>
 
--- a/docs/dashboard/requests/js/requests.js
+++ b/docs/dashboard/requests/js/requests.js
@@ -341,6 +341,13 @@ function initEventListeners() {
 
     DOM.renameBtn.addEventListener('click', () => {
         CopyAppfilter(null, true);
+        DOM.renameOverlay.classList.remove("show")
+        DOM.floatingBtnContainer.classList.toggle("active", state.selectedRows.size)
+    });
+    DOM.keepBtn.addEventListener('click', () => {
+        CopyAppfilter(null, false);
+        DOM.renameOverlay.classList.remove("show")
+        DOM.floatingBtnContainer.classList.toggle("active", state.selectedRows.size)
     });
 
     DOM.tableBody.addEventListener('click', (event) => {
--- a/docs/dashboard/updatable/index.html
+++ b/docs/dashboard/updatable/index.html
@@ -71,6 +71,7 @@
                 <span class="close-button-class">&times;</span>
                 <!-- Preview image will be inserted here -->
                 <img class="preview-image-class" id="preview-image" src="" alt="Preview">
+                <h2 id="preview-image-name"></h2>
             </div>
         </div>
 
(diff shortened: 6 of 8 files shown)


Kaiserdragon2:
Some thoughts:

- use the preview layout from the icon search ore something with a similar design displaying the icon name beneath might live it up a but does look "lost" right now.
- lets add a download button besides the copy button to save the image
- can also add a global download button somewhere (show as floating thing when items selected ?)
- setting button that brings up a popup view to set copy mode with name () or without and possiple other settings regex?

Donnnno:
Great thoughts, and sorry for breaking things right now, I'm slowly fixing everything and learning about the restrucutre magic you've done.

I've added a download button next to the copy button (overwritten your change by doing that).

image

It's still super messy at this point, and I'll try to focus on colors and shapes, but feel free to add changes to this branch too!

Kaiserdragon2:
> Great thoughts, and sorry for breaking things right now, I'm slowly fixing everything and learning about the restrucutre magic you've done.
> 
> I've added a download button next to the copy button (overwritten your change by doing that).
> image
> 
> It's still super messy at this point, and I'll try to focus on colors and shapes, but feel free to add changes to this branch too!

No problem i break my own things all the time. Thanks for reverting my change helped me find a better solution and directly include the download function.

Donnnno:
Oh nice, you already implemented the download button, awesome.
Right now it does save an unopenable PNG instead of the Webp files that they actually are.
image
(file type isn't supported)

Donnnno:
Oh it works now! One last thing I noticed about it: In Chromium based browsers, it works as expected, and in Firefox based browsers, another tab gets opened when downloading.

Also, I'm having trouble getting the arcticons to show when using the 'show matching drawables' button. Could this be related?
image

Kaiserdragon2:
> Oh it works now! One last thing I noticed about it: In Chromium based browsers, it works as expected, and in Firefox based browsers, another tab gets opened when downloading.
> 
> Also, I'm having trouble getting the arcticons to show when using the 'show matching drawables' button. Could this be related?
> image
> 

In Firefox you have to change the standard for webp to save from open its in the settings i was searching like crazy why an window opened every time and it was that setting.

Now i know why you written in your commit that arcticons doesn't work. The appfilter.json doesn't get committed to the repo you need to navigate http://arcticons.com/assets/combined_appfilter.json and save the json for testing in the assets folder

Donnnno:
Ohhh goood to know that!

...and it works now! 
image

Kaiserdragon2:
At the moment when the window gets wider the icon and arcticon get spaced out from each other. Looks a bit weird i think how about we give these two a fixed width?

Donnnno:
Yeah that looks weird, let's do that :)

Donnnno:
I broke everything again, but we now have a more compact design for the regex stuff:
image

Kaiserdragon2:
Lets also create a hovering copy and download button that only shows when items are selected this would also remove the copy button in this row

Donnnno:
Yeah sounds good!

Donnnno:
We have a floating button!
_If only I knew how to center it_


Scherm­afbeelding 2025-12-29 om 16 12 59

Kaiserdragon2:
Should we create a popup view for the category filters ?
Because these buttons and the floating button don't go to well! 
image

Donnnno:
Noticed that too! Or maybe an accordeon menu? Both can work fine I think

Donnnno:
Woow, thanks for fixing everything. It looks and works great already :)

Kaiserdragon2:
An Accordion menu could be better than a popup.

Donnnno:
Especially since we already use multiple popups

Kaiserdragon2:
image
we should try having the floating button always above the footer

Donnnno:
Yeah, that looks kinda weird haha

Donnnno:
It's probably also good to think a bit about the information density. The updatable page has lots of text. Copying only works on the app name, maybe we can hide more info here? 
Also, because of the actions, we only need it to check what's new, not necessarily to use all its possible functionality. 

image

Donnnno:
I think some of our changes broke stuff:
- The floating button stopped showing up.
- Switching back to Show All Entries (from matching drawables) doesn't work anymore.

Donnnno:
Oh, and some more things:
- The X button inside the random button also doesn't work for me or responds with a huge delay (this might also be a local testing thing).
- When selecting a category by pushing its button and then undoing it by pressing it again, the red X clear category button stays active.

Kaiserdragon2:
image
When resized the button width is way to small to click

Kaiserdragon2:
> Oh, and some more things:
> 
>     * The X button inside the random button also doesn't work for me or responds with a huge delay (this might also be a local testing thing).
> 
>     * When selecting a category by pushing its button and then undoing it by pressing it again, the red X clear category button stays active.

Do you use vs code live server or python -m http.server

Donnnno:
> When resized the button width is way to small to click 

Oof, good catch! didn't think of that when I made that button structure. Now the actual button element is hidden all together

> Do you use vs code live server or python -m http.server

Vs code

Donnnno:
Ooo that new design for the updatables page is very nice btw!

Kaiserdragon2:
> > Do you use vs code live server or python -m http.server
> 
> Vs code

had the same problem with vs code 
image
When you open the terminal in vs code 
type `cd ./docs`
followed by `python -m http.server`
that runs way smother at least for me
default port is 8000
http://127.0.0.1:8000/dashboard/requests/

Donnnno:
Damn, that works so much faster! Gonna write that down for myself, thanks!

Kaiserdragon2:
> I think some of our changes broke stuff:
>     * The floating button stopped showing up.

It somehow only is visible when categories are collapsed (hidden)

> * Switching back to Show All Entries (from matching drawables) doesn't work anymore.

don't click the image only the text is a button

Donnnno:
Ah interesting. 

Also, we can make the buttons in the table colorless, that would make the design a bit more boring (I guess?), but also less distracting.

What do you think? 

image

Kaiserdragon2:
i think that would be a good idea
we also could have the links in an "more" menu besides downloads like lawnicons and the have the component displayed under the name
that would help with clarification when a search returns a result

Donnnno:
Sounds good!

Donnnno:
I've fixed the compact button problem, but it might not be a good implementation, so if you have a better solution, feel free to revert it!

Donnnno:
Bringing a bit more color to the dashboard again, lemme know if it's too much

Donnnno:
Nice work with the new popover!

We should probably take a look at the behavior of the copy multiple button. I think the long press is a bit broken now. So maybe split the button up into two buttons:
Copy and rename all, and copy selected ?

Donnnno:
Having the Play Store button directly accessible would be nice instead of hidden in the menu, since 95% of the apps are from the Play Store.

Kaiserdragon2:
image

Donnnno:
> image

The image you sent is just a black rectangle for me. Idk if that's on purpose 🙃

Kaiserdragon2:
it shows the footer in darkmode with dark text

Donnnno:
Ahhh, very cryptic 😆

Kaiserdragon2:
on my screen it was readable but that might be to its huge size

Donnnno:
Or my color blindness 🙃

Donnnno:
Nice improvements! I think we're nearly there with the new dashboard

Kaiserdragon2:
if you want to test the Arcticon column on the updatable page save this in assets https://arcticons.com/assets/package_map.json.
If all entries are filtered out delete combined_appfilter.json

Donnnno:
All entries were filtered out; deleting the combined appfilter did the trick.

Seriously amazing work on the dashboard. The previous version was already so good, but this version is even better! 
Unless I forgot anything, I'd say we're ready to push it to the site. But I'll leave the honor of doing it for you. :ˆ)

PR #3171, opened 2025-12-30
Several fixes and new icons

babyhuehnchen:
icon can be used for all 4 apps as its the same app for different states/areas


babyhuehnchen:
fix for previously closed PR with updated icon

babyhuehnchen:
messed up the naming, need to look into how i can fix that

babyhuehnchen:
naming should be fixed now

Donnnno:
Hi, I took a quick peek at your icons. Some things caught my attention: 

Scherm­afbeelding 2026-01-02 om 13 32 56

- Stroke width: Make sure your stroke width is always 1px/pt 
- Sizing: Icons should fit inside the standard circle and preferably match a similar shape's default size (like a square or rectangle).
- Cluttered icons: Because of the stroke width, you want to make specific adjustments so that your icon won't be too cluttered. This can be done by changing outlined shapes to a single stroke, for example. I've highlighted areas that are relevant.

Other than that, nice work and great to see a new contributor!

babyhuehnchen:
Thanks, I will look into fixing these, thanks for the feedback

babyhuehnchen:
So maybe I'm misunderstanding the guide here https://docs.arcticons.com/contribute/vector-editing-software/vector-asset-creator but I've set the correct stroke width in the app and when i double-checked all of them, it also shows that its the correct one on the final version that I exported, I'll try to look further into it.

EDIT: seems to be an issue with vector asset creator indeed where it doesnt properly apply it in some instances ill go through all icons now and make the adjustments

Donnnno:
Ah, I'm not familiar with Vector Asset Creator, so I can't help with that part. Good that you found the solution, and feel free to ask questions if you need help :)

babyhuehnchen:
Some I'm a bit stuck on, within vector asset creator, lebensretter.svg for exaple looks perfectly fine with the stroke width set to 1 and by a longshot not as extreme as on the ones you linked. Is there an easier way to fix this cause at this point i think its the app being buggy or me simply overlooking something stupidly little

babyhuehnchen:
Seems to be the app indeed, checking it out in https://editsvgcode.com/ it sets stroke with in a weird way. When I set the stroke width to 10 in the app, in reality, it sets it to 0,48. I'll need to look into learning another app to fix in than or fix the icons first and than re-import them into vector asset creator afterwards

babyhuehnchen:
further, i just saw that there are icons in the repo that arent mine, ill remove them with the next PR, no clue where they came from, or do i need to keep them in?

babyhuehnchen:
I have attempted fix everything as per your request. By default, Vector Asset Creator on Android messes with the stroke width, setting it to 21 however sets the value so close to 1.000 that fixing it after in notepad++ changes it so minimally that its indistinguishable to the human eye

Donnnno:
I've changed some bits, but overal good work!

image

Oh when you're adding appfilter entries to existing apps, please add them after the existing ones in the appfilter.

PR #3190, opened 2026-01-11
Gradle and java Improvments

Kaiserdragon2:
@Donnnno I reworked the preparehelper to improve performance and clean up the code.
Please test and see if everything still works as expected


Donnnno:
Ohhh, that's great, will take a look when I merge every other PR :)

Donnnno:
Only did one test so far, and it went smoothly. Haven't had time to do some more testing, but hopefully I will find some time this weekend. :)

Kaiserdragon2:
> Only did one test so far, and it went smoothly. Haven't had time to do some more testing, but hopefully I will find some time this weekend. :)

Found some bugs while transitioning to AGP 9  The Icons wouldn't be correctly counted.

Donnnno:
The run checks action is a bit confusing now because the first thing you get is this:
image

Instead of the overview:

image

Not a big deal tho, because it does the same as before.

Donnnno:
This looks good! Maybe add a green check emoji here to emphasize it? 
```
> Task :preparehelper:com.donnnno.arcticons.helper.PrepareRelease.main()
Starting task: checkonly
No violations found. All checks passed!

```

Also, it might be worth it to look at this?
image

Kaiserdragon2:
The checkmark isn't displayed in the console. Because of that i set it to error out if checks fail to make it more obvious that something is wrong. I am not sure if it is possible to show the overview instead of the other failure code.

The jetifyer thing is dependend on candybar. If it is fixed there we can disable it also it is nothing to worry about for now.

Donnnno:
Ah ignore my comments then, I'll test it on my MacBook later to see if it throws any errors and then we can merge it :)

Donnnno:
It works like a charm! 
Some nice-to-haves I can think of, but do tell if it's much work.
For the release image:
- Change the background color to #050714 (matching the request dashboard's background color)
- Add a text above the icons with something like: Arcticons 14.0.0 — xxx new icons. 

But that can be something for the future too.

Feel free to merge it when you think it's ready. Amazing work!

PR #3197, opened 2026-01-15
add metrolist

ghost:


Donnnno:
Nice work, I corrected the stroke width and made some tweaks to the shape

PR #3206, opened 2026-01-18
add word maker

ghost:
right one this time


Donnnno:
The lines were a bit thin so had to fix that :)

image

PR #3166, opened 2025-12-28
Add 40+ app icons

Npepperlinux:
Add app icons:
- AstroPay
- Daiichi
- Elgato Stream Deck Mobile
- e签宝 (as "eSign")
- 方正证券 (as "Founder Securities")
- はてなブックマーク (as "Hatena Bookmark")
- HSBC PayMe
- iD Mobile
- 日本通信アプリ (as "Japan Communications")
- カクヨムビューワー (as "Kaku Yomu Viewer")
- KB스타뱅킹 (as "KB Star Banking")
- LevTech
- LINE Manga
- looopでんき (as "looop Denki")
- マナビジョン (as "ManaVision")
- Mind Cafe
- Movie Creator
- my ac
- NIPPON Rent-A-Car
- OpenCloud
- OpenEvidence
- Paypo
- Pushin Pay
- qBitConnect
- Rakutenおまとめ (as "Rakuten Omatome")
- Читай-город (as "Read-City")
- Shinkansen smartEX
- Soup Stock TOKYO
- サミット (as "Summit Store")
- サンデーうぇぶり (as "Sunday Webry")
- Timee
- viviON
- 352Life
- and ST
- Ao Oni Series
  - 1, 2, 3, X
- d Card
- food2you
- Food You
- Lynk Taxies
- Orb
- SmartLife-SmartHome
- STid
- TP-LINK IOT
- Xiaomi Glass

PR #3235, opened 2026-02-03
Fix 2 icons & Remove duplicated icon

Npepperlinux:
Fix "d card" duplication
- My fix for mistake #3166 seems to have caused another mistake.

Fix "PrivacyShade" icon style
- Two CSS path classes existed, and one was not being displayed, so they were merged into one.

Remove duplicated icon
- Duplicated with "privacyguide" & "privacy_guides"


Donnnno:
Ah thanks for the fixes!

PR #3265, opened 2026-02-19
Adds new app: Reon Pocket

phillip-bruno:
Hopefully the SVG is okay, I'm really bad at creating these types of things so let me know if I ended up missing a design requirement from the guidelines.


babyhuehnchen:
You should outline the rectangle rather than completely filling it 😃 other than that, all good as far as i can see.

and don't worry, its a learning curve, my first icon took me multiple tries and hours to get right 😉

Donnnno:
I noticed that the icons size was huge, so had to fix that :)

phillip-bruno:
Oh, I see that sizing error now 🤦 Thank you for fixing that!

PR #3266, opened 2026-02-19
add 5 new app icons

tsehla42:
First contribution, feel a bit nervous


Donnnno:
Nice work on the icons!
I've resized them a bit and modified the cloud one to look more clean.

image

Donnnno:
I've also renamed the icon with Cyrillic characters; non-Western characters most likely crash the app, so we're only using standard Latin names.

PR #3268, opened 2026-02-21
Added icon for Nomad Esim

orbitaldr0p:


Npepperlinux:
The icon with the drawable name `nomad` [already exists](https://arcticons.com/arcticons/search/?icon=nomad), so you need to change both the drawable name and the icon file name to `nomad_esim`.

PR #3269, opened 2026-02-24
[FIX] App_Deny-List grammar & [ADD] Another Revanced Manager activity

NeoNyaa:
Adjusted grammar for snatcher entry, did change date modified but that can be undone if thats only to be changed when adding an app


Donnnno:
thank you!

PR #3270, opened 2026-02-26
feat: Add 12 icons

troxor:


Donnnno:
Nice work, just made some sizing tweaks; other than that they're good!
image

PR #3272, opened 2026-02-28
350+appfilter lines

Donnnno:


TotallyAvailable:
The appfilter and crowdsource workflows still need their token permissions defined (currently the default write-all) for the 'check-pr' part @Donnnno , @Kaiserdragon2 

Ideally by starting every new workflow file with 

```
permissions: {}
```
or a less strict 
```
permissions:
 contents:read
```
at the top before any further fine-tuning, outside the specific chains...given how easy it is to forget.

Set 'persist-credentials' to 'false' unless ***absolutely necessary !***

Consider the risks associated with scheduled workflows.

Consider the inclusion of SECURITY.md and 'CODEOWNERS'.

Also worth considering avoiding the use of 'pull_request_target' at all cost unless absolutely necessary, ***especially*** combined with 'actions/checkout'. On top of setting the 'Secret Source:' to none instead of Actions for the 'Build Pages' workflow unless required. Again several (big) repos paid big time for this one recently.

Also no PR ever should be able to touch (at least) the workflow folder unless manually triggered (!) given how little interest/care there appears to be coming from GitHub even after shit hit the fan several times as mentioned above...

Checksums in the readme (visible change through commit history) to prevent potential "silent" malicious changes to the release note and assets if immutable releases aren't an option or wanted might not hurt either.

Some devs even added an automated Virustotal check into the release workflow, which would obviously work nicely combined with the readme checksums. (Just a couple more thoughts though. Benefits compared to the current way of doing it debatable of course)

PR #3288, opened 2026-03-10
Added icons for Antiyoy Online and Endless ATC.

gabrielh-so:
Hello,
Here are a couple of missing icons. It fixes what I think is the pack autodetecting a main view, resulting in a kawaii looking Endless ATC. Also the Antiyoy Online app icon is just a mirror image of the offline version, matching the actual app icon pair. Hopefully these are okay; it's my first time contributing.
Thank you for the great icon pack!


Donnnno:
Nice work!

PR #3311, opened 2026-03-18
Add cathay pacific + sg airlines icons

J0e-Quan:
also added extra activities for maxis appfilter


Donnnno:
Did some shape cleanup and fixed the line-width:
image

PR #3312, opened 2026-03-18
9 new icons and appfilter fixes

babyhuehnchen:
added ESTKme and WoGibtsWas


Donnnno:
Thanks again for your work, I made some fixes, mainly changing lines to have round caps and corners, make the circle dots the right size and some other cleanup.

PR #3314, opened 2026-03-19
k9 will soon become thunderbird :-(

aloxe:


aloxe:
There is no icon, just a k9 updates that changes its appfilter

Donnnno:
duplicate of #3310 ?

aloxe:
> duplicate of #3310 ?

As far as I understand the change in #3310 is related to people having thunderbird that updated to a new thunderbird. The one I am pushing is for people having K9 that are updating to thunderbird but with the same repo as before (com.fsck.k9)

So not a duplicate. The icon already exists though.

PR #3315, opened 2026-03-20
add icon support for a few apps using existing icons

thomas-sim:
the icon request dashboard is such a great tool ! it's very "fun" and easy to use it for quick contribs like this


Donnnno:
Please add the entries below the existing ones next time :)
Thanks for the new additions!

PR #3320, opened 2026-03-25
added unite students and morphe manager icons

orbitaldr0p:


Donnnno:
Morphe was also added with another PR.

Unite Students didn't match our style guides, so I remade that one.

Scherm­afbeelding 2026-04-20 om 13 52 21

PR #3337, opened 2026-04-04
Adding 1 app

Frenchfrench:
This is my first icon : for the app AELF. I resized it to take the place recommended by the guide (37 px rectangle ∩ 43px circle). Don't know if it's actually the right thing to do

Screenshot From 2026-04-04 11-41-20
aelf

### DIFF
--- a/newicons/aelf.svg
+++ b/newicons/aelf.svg
@@ -1,145 +1,15 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<!-- Created with Inkscape (http://www.inkscape.org/) -->
-
-<svg
-   width="12.7mm"
-   height="12.7mm"
-   viewBox="0 0 12.7 12.7"
-   version="1.1"
-   id="svg1"
-   inkscape:version="1.4.3 (0d15f75042, 2025-12-25)"
-   sodipodi:docname="arcticons-aelf.svg"
-   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
-   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
-   xmlns="http://www.w3.org/2000/svg"
-   xmlns:svg="http://www.w3.org/2000/svg">
-  <sodipodi:namedview
-     id="namedview1"
-     pagecolor="#ffffff"
-     bordercolor="#000000"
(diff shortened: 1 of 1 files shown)


aloxe:
You are almost there @Frenchfrench size and stroke look good except that I think we should send it with a white stroke.

Also if the size is correct, it appears smaller than other icons (see [this PR for example](https://github.com/Arcticons-Team/Arcticons/pull/3325/changes)) the viewbox should be `viewBox="0 0 48 48"`. Since you are using inkscape, you should try to open one of the templates, remove the existing lines and copy your image into the box. The viewbox should then be correct.

One last thing I am not fond of the tail at the bottom of a long line. It doesn't look like a bookmark ribbon. I would remove it but this is up to you ;-)

Frenchfrench:
Hi @aloxe,
thanks for your kind help !
I think I fixed the svg as you described. I've left the tail though, as I think it is an important part of the logo (simbolizing the bookmark in a bible). Thanks again, heave a nice day ! and tell me if there is anything else to change ;-)

Frenchfrench:
What's strange is that github says the new picture is 300px*300px, 

image


but the file says 48*48 :

```<?xml version="1.0" encoding="UTF-8"?>
<svg id="a" version="1.1" viewBox="0 0 48 48" xmlns="http://www.w3.org/2000/svg">
 <defs>
  <style>.b{fill:none;stroke:#fff;stroke-linecap:round;stroke-linejoin:round;}</style>
 </defs>
 <g transform="matrix(3.6035 0 0 3.6035 .77902 2.2343)" fill="none" stroke="#fff" stroke-linecap="round" stroke-linejoin="round" stroke-width=".2775">
  <path d="m9.5222 10.678 0.28816-0.61671"/>
  <path d="m5.5209 1.045h0.94233l3.3233 9.0136 0.61671 0.28816"/>
  <path d="m6.1367 2.2835-2.1972 5.2872"/>
  <path d="m7.336 5.5697-0.25218-0.6645-1.9492-0.0045832"/>
  <path d="m3.4217 7.5707h0.98884"/>
  <path d="m6.02 1.9553 0.31729 0.89188"/>
  <path d="m7.0029 4.6779 0.31729 0.89188"/>
 </g>
</svg>```

Donnnno:
Nice work on the icon!

PR #3352, opened 2026-04-17
Added 12 icons

hugarburdur:
Added icons for following apps:
Total Elec & Gaz 
Hanafuda Koi Koi Dojo
ISS Vanguard
Kula
Netto
Oeni
Panorama Mobile
Ringotan
Roole Map
Ruupa
Satori Reader
U-Boot The Board Game


Donnnno:
Nice work on the icons, they look really good! 
I've changed some minor things:

Scherm­afbeelding 2026-04-27 om 12 25 27

As you can see, I've removed some box shapes to give the icons in them some more room to shine :)
I've also noticed that you used some clipping masks. We rather not use those so I removed those too

PR #3364, opened 2026-04-20
Add updated icons with some fixes

Npepperlinux:
I made several adjustments to the existing icons:
- Since they were composed of filled paths, I redrew the line paths
- Fixed an issue where some lines extended beyond the border when CSS styles were applied correctly
- Only replaced `height, width` attributes with `viewBox`


Donnnno:
Thanks for the fixes :)

PR #3369, opened 2026-04-25
Add generic icons for Pill Reminder, Tooth Brushing Timer

KatSteinke:
While these two types of apps aren't on the explicit list of generic apps in the contribution guide, the ones I've added have icons that match/can be simplified to existing generic icons (pill for Pill Reminder, tooth emoji for Tooth Brushing Timer). Hope that's all right?


Donnnno:
Looks good!

PR #3383, opened 2026-04-29
Added new Stadtradeln activity name

alchemiker:
Since the recent update of the Stadtradeln app broke the icon due to a renaming of the package from `StartupActivity` to `MainActivity`, here a fix for that. And I also deleted a NBSP (non-breaking space?) displayed in my editor.

So a minor fix this time, no new icons.


Donnnno:
nice!

PR #3392, opened 2026-05-03
Update gradle

Donnnno:
From the codeQL action


Kaiserdragon2:
I would leave the scripts as they are. They probably will brake with these changes. Also i think all of these aren't used anymore.

Kaiserdragon2 (changes requested):
Leave script's as they are. Gradle and manifest changes are fine.

Donnnno:
Thanks, will do! Maybe we should remove them if we're not using them anyways?

Kaiserdragon2:
I always wanted to clean the scripts up a bit but never came to it. Yes indeed we should remove the old ones.

yachnevskijea-beep (commented):
**[]()



**

PR #3410, opened 2026-05-22
29 icons added

Karithina:
Added 29 icons including a few aussie specific things like qantas, qld rego check etc.


Npepperlinux:
I have two concerns:
1. `iitcmobile` icon is exists and `iitc` icon is duplicated with that.
2. The path objects for the `bunnings` and `snap_send_solve` icons appear to be too small.

Karithina:
@Npepperlinux Thank you, valid concerns. I've removed my duplicate, linked iitc to the existing iitcmobile app. Resized quite a few icons because I'd actually misunderstood the templates as a safe zone rather than the suggested closeness to the edge, so many were a bit small. Should be better now.

Donnnno:
Hey! Amazing work on your first contribution, I've looked at the icons and they're great!

Some things that I changed:
- Shapes around text only icons
- Removed some unnecicary text elements

I also was doubting a bit on what to do with the Snap Send Solve & TSV Parking icons because they're a too text heavy for my taste.


Scherm­afbeelding 2026-06-19 om 21 21 08

Karithina:
@Donnnno Ohh the squares and circles make sense now for balancing text icons, great reductions! Yeah those two could be simplified to likely a generic camera icon and a ticket icon, but you've likely done similar by now while merging. Much appreciate you showing how they looked before merge, gives me a better idea of the style for the next branch :)

PR #3411, opened 2026-05-22
Add Areada icon

sp72921:
I created an icon for Areada PDF Reader.


Donnnno:
Fixed the icon to be in line with the Arcticons:

image

image

PR #3420, opened 2026-06-12
added msplit

faezzi:
This is my initial attempt at contributing. If all goes well, I might contribute a few other icons I' m missing.

edit: If needed, msplit can be found at https://github.com/Zingler/Msplit


Donnnno:
Noticed that there were two lines, so cleaned up the svg. 

image

image

PR #3432, opened 2026-06-19
Apply Slack icon

Liquid-Aristocracy:
Slack is using another package name, so I updated appfilter.xml to include it.


Donnnno:
thanks :)

PR #3444, opened 2026-07-05
Add missing Essentials icon

ImenaOphelia:
Present in appfilter.xml of #3443.


Donnnno:
Found some issues with this icon:
- a bunch of clipping masks
- sizing was too large
- wrong stroke-width.

So I went ahead to fix it, giving it a consistent stroke width makes it also really cluttered, what do you think?

Scherm­afbeelding 2026-07-05 om 12 50 29

Scherm­afbeelding 2026-07-05 om 12 53 00

ImenaOphelia:
> Found some issues with this icon:
> 
> * a bunch of clipping masks
> * sizing was too large
> * wrong stroke-width.
> 
> So I went ahead to fix it, giving it a consistent stroke width makes it also really cluttered, what do you think?
> 
> Scherm­afbeelding 2026-07-05 om 12 50 29 Scherm­afbeelding 2026-07-05 om 12 53 00

Hello, what I did with this icon was changing details like stroke width in straight-from-the-source icon SVG (https://github.com/sameerasw/essentials). I think preserving the app's author idea when making it compatible with basic Arcticons guidelines like I tried to do would be the best option. The thicker stroke width version looks cluttered indeed.

Issue #3501, opened 2026-08-08
[CI/CD] Add github actions for nightly/weekly builds


LuminarLeaf:
Since there can be a large window between releases quite a lot of apps which changed the activity go unresolved during that phase and this causes loss of quality (imo) in the icon pack.

I request a nightly or weekly build using github's pre-release feature for people who want quick updates and it would also help in testing the pack more regularly as developers / "beta testers" can test the changes implemented more quickly and if required polish the app / logos before the release.

Donnnno:
Interesting idea! 
I have no idea what a good way would be to implement this, but if someone has a good practical solution, I would happily merge it :)

Donnnno:
One important thing should be that the /newicons folder should not get cleared for beta releases

Kaiserdragon2:
Merged #3510 it works only problem is to download the apk the user needs to be logged in.

Donnnno:
@LuminarLeaf we now have weekly builds! :)

PR #3510, opened 2026-08-12
add BuildNightly.yml to test

kitters-kat:
Attempting to naively address #3501 by recycling code from signed debug release. It *seems* to work for me up to the point where gradle gets upset because I didn't add a key to sign with. Currently set to run daily at noon New York time, but that is easily configurable to run on a different schedule or whenever a commit is pushed. There might still be the matter of versioning and activity name to address (i.e. `Arcticons-normal-nightly-2026-08-12.apk` and `com.donnnno.arcticons.nightly`).

### DIFF
--- a/.github/workflows/BuildNightly.yml
+++ b/.github/workflows/BuildNightly.yml
@@ -1,8 +1,8 @@
 name: Build Nightly
 on:
   schedule:
-    - cron: '00 12 * * 0-6'
-      timezone: "America/New_York"
+    - cron: '00 12 * * 6'
+      timezone: "Europe/Amsterdam"
   workflow_dispatch:
 
 permissions:
--- a/.github/workflows/BuildNightly.yml
+++ b/.github/workflows/BuildNightly.yml
@@ -1,0 +1,60 @@
+name: Build Nightly
+on:
+  schedule:
+    - cron: '00 12 * * 0-6'
+      timezone: "America/New_York"
+  workflow_dispatch:
+
+permissions:
+  actions: none
+  attestations: none
+  checks: none
+  contents: read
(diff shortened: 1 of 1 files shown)


Donnnno:
Super interesting! Cool that you made an implementation.

Maybe daily is a bit much - since there is not much happening daily, but weekly could be very useful!

The signing should not be an issue since all keys are stored inside the secrets. 

@Kaiserdragon2 could you take a look and check for things that I might have missed? Thanks :)

kitters-kat:
Mostly just a copy and paste from https://github.com/Arcticons-Team/Arcticons/blob/main/.github/workflows/BuildRelease.yml, looking up some github workflow documentation, and a few trial runs to debug on my part. Probably still issues, but should be a start at least.

Kaiserdragon2:
I will have a look.

Donnnno:
I was just thinking: It would only be effective if PRs are merged on a weekly basis.
As for the past months, there were moments that it was not possible for me to merge all PRs, giving me sometimes huge backlogs to work with.
So in reality, this would be the most useful for appfilter updates probably.

Kaiserdragon2:
> I was just thinking: It would only be effective if PRs are merged on a weekly basis. As for the past months, there were moments that it was not possible for me to merge all PRs, giving me sometimes huge backlogs to work with. So in reality, this would be the most useful for appfilter updates probably.

That is correct but even that could be enough sometimes

Donnnno:
To check, some last questiona on my mind: 
- should a version number be bumped with weekly releases?
- are they marked as pre releases in gh?
- will they be published to fdroid/ Google play?

PR #3499, opened 2026-08-03
Added Icons for various apps

CouchhhausenL:
airplay_server
arrow_escape
backpack_brawl
baluchon
brick_blast
calendula
edge_roll
esa_impact
geminiassist
markleaf
materializr
mj_pdf
remember
scorehub
solitaire
square_on_top
swisstransfer
unblock_jam
universe
volksvertreter
xnotes

PR #3511, opened 2026-08-13
Added Icons for various apps - #3499

Donnnno:


CouchhhausenL:
Hey short question, when will the icons be in the app? How long does it usually take?

Donnnno:
I aim for monthly releases, but it has been difficult to pull off this year

PR #3516, opened 2026-08-15
added flclash icon

orbitaldr0p:


Donnnno:
Did some shape cleanup:

Screenshot 2026-08-15 at 15 46 40

PR #3543, opened 2026-08-25
change readme format slightly, move badges

kitters-kat:
Wanted to see if the badges might look nicer in rows under the logo. Thoughts?


Donnnno:
Good one!

Issue #1648, opened 2023-08-11
Off-Center Icons


raycadle:
First off, I just want to say thank you for the amazing icon pack. Made my Pixel so much nicer to look at. Keep up the great work!

On a different note, however, I noticed a minor bug in some icons. A couple FOSS apps — Newpipe and Retro Music Player — have their icons off-centered and it's really tripping out my OCD. I'm using Lawnchair 2 as launcher and never had the icons off-centered with the stock icons.

![Screenshot_20230811-085906~2](https://github.com/Donnnno/Arcticons/assets/93884664/6fa5e82e-4dd5-4a40-9701-e4a2e1d7d1b3)
![Screenshot_20230811-085941~2](https://github.com/Donnnno/Arcticons/assets/93884664/0f4f5752-b4e5-4d61-87cc-7d8c87600e4f)

As you may notice, Retro isn't severely off-centered, however, Newpipe is very noticable. I don't want to rush anyone to fix this. I just didn't see a related issue open so I thought I would bring it up, in hopes that it silently gets fixed in a future update.

With that said, thanks a bunch and have a great rest of your day! Blessings to you and yours.

Donnnno:
Ah, that's a good one. I'm gonna take a look at it. From what I can tell, the triangular icons are probably aligned in the middle. But they should be bumped a little to the right so that it's optically in the middle.

dgcampbe:
Perhaps we could find a way to make a script automatically align icons to ensure that all of them are mathematically perfectly centered. When I create icons, I am not always sure exactly what point should be in the center, so I kinda eyeball it.

Donnnno:
In this case, you actually want it to be slightly off centered hahah, so a script cant do that. Also, all shapes have to be grouped (or made a compound path) properly.

Donnnno:
I also think this is an Arcticons You specific problem, it never bothered me with the regular versions, but it makes sense for the You version to be modified a bit.

Donnnno:
This is a difficult one because of all the different sizes of the adaptive shapes.

wwwwwwari:
I wonder if this only happens with triangular shaped icons?

Donnnno:
Yes

Donnnno:
![image](https://github.com/Donnnno/Arcticons/assets/31142286/e90f3c63-e118-4687-af70-a4cf5b3c46b3)

this is better already

KR4U7H3R:
[...]

PR #3547, opened 2026-08-26
Added 1 new icon, 1 new map, 1 new (re)map +fixed said remapped icon

avant-chrysolite:
Added new icon for:
- Dusklight

Mapped an existing icon for:
- BB Racing (Arcade)

Fixed/improved existing icon for:
- Retro Music

<details>
<summary>Show reason</summary>
<br>annotation showcasing incorrect vertical alignment of an existing icon
Original <i>retroplayer.svg</i> seems to have an offset of about ~3px on the x-axis.<br>
It also used multiple paths for approximating curves and had minor parallel discrepencies.<br>
They don't seem intentional, so they're replaced with more precise geometry.<br clear="both">
</details>

<b>Re-</b>mapped an existing icon for:
- Metro Music Player

<details>
<summary>Show reason</summary>
This is more of a personal opinion, but the existing icon does not do a good job of representing said music player. Especially when both <i>Metro</i> and <i>Retro</i> share the same icon natively (fork).<br><br>
No hate towards the creator at all :3
</details>


avant-chrysolite:
oop, it seems GH doesn't automatically update git branches for active PRs

Donnnno:
> Fixed/improved existing icon for:

This is an interesting one that has to do with optical sizing.
Triangle shapes are hell when it comes to that; the current version is aligned from the middle. Which means that if you compare it to circular shapes, for example, it feels in balance. 
The middle alignment option is a compromise, of course, but it's there because the majority of launchers work with a grid instead of a list. 

We talked about this before in this issue: #1648

avant-chrysolite:
> This is an interesting one that has to do with optical sizing.
> Triangle shapes are hell when it comes to that; the current version is aligned from the middle. Which means that if you compare it to circular shapes, for example, it feels in balance.
> The middle alignment option is a compromise, of course, but it's there because the majority of launchers work with a grid instead of a list.
> 
> We talked about this before in this issue: #1648

I initially had a feeling that the offset was intended afterall.
I don't have experience using circular icons so viewing the screenshots in that issue made it easier to spot haha. It was a selfish change on my part, apologies!

I added back the offset. The path logic remains unchanged so it could be a *teeny* bit more useful to use this, instead of the current svg which uses cubic Bezier for curve approximations (as opposed to just 3 arcs).
It's more of a precision nitpick (selfish, again) than anything, but they look basically identical regardless. :)

Additionally, should we map "Metro" as `metroplayer` or, `retroplayer` like the many forks of "Retro" as well?
I was curious since while [here](https://github.com/Arcticons-Team/Arcticons/pull/3547/changes/f54ffe4635985cbd01c2ebb2a24c12871fd60efe#diff-f6276b41e7f583990e94ba5458fb136f211f7f4a89a5f595ecfd02de76527c83R40149) it has its own unique entry and corresponding drawable name, [here](https://github.com/Arcticons-Team/Arcticons/pull/3547/changes/f54ffe4635985cbd01c2ebb2a24c12871fd60efe#diff-f6276b41e7f583990e94ba5458fb136f211f7f4a89a5f595ecfd02de76527c83R55279) one of its other activities were mapped as `retroplayer` with various forks.

Donnnno:
Haha no worries.
I found some more mappings containting retroplayer and added those to the icon as well

PR #3557, opened 2026-08-30
improve Nightly build action

kitters-kat:
Added prepare release helper and found a package for making releases to the action
Also added a nightly application id suffix

idk if it works

#3501


Kaiserdragon2:
instead of this package use the integrated `gh` command to create releases 

```
# Publish as a GitHub Prerelease
          gh release create "$TAG_NAME" \
            --prerelease \
            --title "$RELEASE_TITLE" \
            --notes "$CHANGELOG" \
            --target "$GITHUB_SHA" \
            app/build/outputs/apk/normal/nightly/*.apk \
            app/build/outputs/apk/black/nightly/*.apk \
            app/build/outputs/apk/dayNight/nightly/*.apk \
            app/build/outputs/apk/you/nightly/*.apk
```

this snippet uses vars for title and change log  but something hard coded for the nightly build might be fine.
i am not sure if it would be possible to use a fixed tag and overwrite it each time. 

Your ToDo assumption is correct just change Release to Nightly

kitters-kat:
Haha the prepare release helper github action works!

Donnnno:
Since this is not my expertise, let me know if it's ready to merge ^^

Kaiserdragon2:
Do we need the preparehelper action. Until now this step is run locally and i would keep it this way because it runs some checks and fails the run if these aren't finished without errors. That would make the workflow more complex to run this step in ci. because every time something is fixed it needs to get pushed then the ci has to run. it could potentially have another problem and the cycle would continue. when running locally you run it until everything is ready and then push it.

  If i am not mistaken the TagName variable is not set. If overwriting works we could just name the tag nightly.

kitters-kat:
Ok, I can move it to be a step run for the nightly build, but not pushed to the actual repo.

Kaiserdragon2:
It already is run in your nightly action

Donnnno:
That would be great yeah.
The local prepare action gives me the control to test it locally, which I like very much.

kitters-kat:
ok, should be ready to merge and test the functionality for creating a github prerelease

PR #3566, opened 2026-09-02
Merge the changes made previously on Codeberg

kitters-kat:
A few additions to appfilter and a few workflows that may or may not be useful, plus a punch of merge conflicts I looked over and decided to discard. Since the commits are still there, they should always be able to be recovered later, anyway. You can ignore the zillion commits and just look at the net changes.


Kaiserdragon2:
These workflows or at least what they do exist already in here. If there are some appfilter changes i would only transplant these. Without the history.

kitters-kat:
Alright, I'll ditch the history and workflows

kitters-kat:
Actually checked every appfilter entry and no actual additions would be made, all dupes. My bad

PR #3575, opened 2026-09-05
Add WizeStream icon

deadendpl:
Added a [WizeStream](https://github.com/wizdom13/WizeStream) icon.  
The lines there might not be as straight as the original icon because it has a weird shape, so that may need adjustments.

### DIFF
--- a/newicons/appfilter.xml
+++ b/newicons/appfilter.xml
@@ -78316,4 +78316,7 @@
 	<!-- 献血Web会員サービス ラブラッド -->
 	<item component="ComponentInfo{jp.kenketsu.blood_donation_card/com.example.blood_donor_card.MainActivity}" drawable="cross"/>
 
+    <!-- WizeStream -->
+    <item component="ComponentInfo{org.wisso.newpipematerial/org.schabi.newpipe.MainActivity}" drawable="wizestream"/>
+
 </resources>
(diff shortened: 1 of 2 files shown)


Donnnno:
Hey! You can ignore this part:
```
[ DUPLICATE ENTRY ]
  -> appfilter.xml: ComponentInfo{com.movtery.zalithlauncher/com.movtery.zalithlauncher.ui.activity.SplashActivity}
  -> appfilter.xml: ComponentInfo{com.movtery.zalithlauncher.v2/com.movtery.zalithlauncher.ui.activities.SplashActivity}
```

deadendpl:
I had the stroke width in inkscape set to 1 px.  
Should I set the stroke width in SVG file directly to 1?

Kaiserdragon2:
The problem comes from this element `transform="matrix(`. This happen when you group things and then scale them.
To fix this:
1. Edit -> XML Editor -> select the group -> delete the transform element
2. Ungroup everything
3. Object -> Transform -> Check "Apply to each object seperatly"
4. Rescale everything
5. Check if stroke-width is set correct

deadendpl:
Thanks for the help! I think I fixed the icon now.

Donnnno:
Nice work on your first icon! 
All looks good :)

PR #3577, opened 2026-09-06
Added check to find transform matrix elements 

Kaiserdragon2:
These always trip up the stroke width check.


Donnnno:
Woah, you found the thing that has been bugging me for years now 😅 
Great that we can finally pinpoint it for future icons :)

PR #3586, opened 2026-09-08
Update & Cleanup icons

Npepperlinux:
Update icons
- Vandle Card
  - recently redesigned
- Homebase Resident
  - I was concerned that the hexagon looked more squashed than the one shown on the Play Store.

Cleanup icon:
- Phira
  - Since the path has a very large number of nodes, here is a suggestion for smoothing the path and reducing overlaps.
Screenshot from 2026-09-08 15-37-32


Donnnno:
Thanks for these fixes :)

PR #3595, opened 2026-09-14
Icons for VARTA apps added

NK308:


Donnnno:
Nice work on the icons! I rounded the corners on the square icon to make it fit more with the rest

PR #3601, opened 2026-09-19
Added 6 icons and updated Booming Music

Elaborendum:


Donnnno:
Looks like you're having trouble removing the transform elements, these steps should help:

- Edit -> XML Editor -> select the group -> delete the transform element
- Ungroup everything
- Object -> Transform -> Check "Apply to each object seperatly"
- Rescale everything
- Check if stroke-width is set correctly.

PR #3605, opened 2026-09-19
Update 

Donnnno:


Donnnno:
interesting to see that the bot thinks all the SVGs are missing @Kaiserdragon2 
anyways, will be ignoring it because I know they're there :-)

